PR #189, opened 2015-11-29
Fix QWaitCondition warning by separate thread

teamikl:
## Summary of this commit changes:

- Move inner classes in cpp files to header file.
  Because, QObject require to include "moc" file.
- Separate worker class.
- remove the destructor

----

- https://blog.qt.io/blog/2010/06/17/youre-doing-it-wrong/
  This article explains the confuse part of QThread and  new Qt5 thread

----

This is a WIP PR (Work In Progress Pull Request) practice with real patch.

Note to Assignee
Please do not merge until those tasks are consumed.
When the time to go merge, I will remove [WIP] on the title.

- [x] thread context
  addSamplesToEncode method is a just trigger to input data. but may be blocked.
  does it need to be thread context by QMetaObject::invokeMethod, or can be in main thread.
  I have a commit on local (not pushed to remote yet)

- [x] I am not testing the side effects by this changes.

- [x] Is it safe to remove the destructor?
      Not sure at this moment, so need to debug more info.

      thread and worker both inherits QObject, they can be put on QObject heap
      usually no need manual delete code. if there is same life span object,
      he can become the parent(parent node of the heap, not super-class)

- [x] @elieserdejesus you can put Milestone to this PR
      I choose this topic because this was commented in TODO
      and priority can be low, users may not notice this warning.
      not next version but near future (not so big task).

     I have sub-topic about version. Qt has version <major>.<minor>.<patch>
     when Jamtaba increment the middle version?

- [x] review codes, small matters

     e.g. I move class to header, now the NinjamController.h has several classes
     they should be separate files, but I was not sure where should I put.
     namespace and path of file.

     this project does not have coding style rule. (I mean here, not documented on wiki)
     I believe following Qt's official is the best way, because Qt has pre-process by qmake.

     If you have own idea/experience,
     but It is hard task to write the document from scratch, so following standard is easy way,
     use one of standard style and customize, write the difference as project local rule.

  - QtCreator has coding style support, she looks style for code analysis.
    you don't need to care this writing time, just auto format before commit.
  - ambiguous is confuse for contributors.
    having coding guide make them feel easy to make a patch.

- [x] no Unit tests
  This is a big topic, this PR will not have unit tests.
  Maybe small example code to show what was problem and what this patch fixed.

  @elieserdejesus we had talked about CI before. I know this is also has initial cost, tho.
  if JamTaba use open source license with public repo
  it's a chance to use those service.
  I am not sure about VST SDK which another license.
  but at least some parts can be test as sub modules.

I know this post has several sub topics which not usually talked in PR.
these are should discuss other thread or chat, this time I just leave as the memo.

Now this is a beginning stage, I may put more tasks when I noticed.

### DIFF
--- a/Client/src/NinjamController.cpp
+++ b/Client/src/NinjamController.cpp
@@ -242,7 +242,11 @@ void NinjamController::process(const Audio::SamplesBuffer &in, Audio::SamplesBuf
                             mainController->mixGroupedInputs(groupIndex, inputMixBuffer);
 
                             //encoding is running in another thread to avoid slow down the audio thread
-                            encodingWorker->addSamplesToEncode( inputMixBuffer, groupIndex, isFirstPart, isLastPart);
+                            QMetaObject::invokeMethod(encodingWorker, "addSamplesToEncode", Qt::QueuedConnection,
+                                                      Q_ARG(const Audio::SamplesBuffer, inputMixBuffer),
+                                                      Q_ARG(quint8, groupIndex),
+                                                      Q_ARG(bool, isFirstPart),
+                                                      Q_ARG(bool, isLastPart));
                         }
                     }
                 }
--- a/Client/src/NinjamController.cpp
+++ b/Client/src/NinjamController.cpp
@@ -21,94 +21,62 @@
 
 #include "audio/samplesbufferrecorder.h"
 #include "Utils.h"
-#include <QWaitCondition>
 #include "log/logging.h"
 
 
 using namespace Controller;
 
 //+++++++++++++  ENCODING THREAD  +++++++++++++
-class NinjamController::EncodingThread : public QThread{//TODO: use better thread approach, avoid inheritance form QThread
-public:
-    EncodingThread(NinjamController* controller)
-        :stopRequested(false), controller(controller){
-        qCDebug(jtNinjamCore) << "Starting Encoding Thread";
-        start();
+
+EncodingWorker::EncodingWorker(NinjamController *controller, QObject *parent)
+    : QObject(parent), stopRequested(false), controller(controller)
+{
+}
+
+void EncodingWorker::addSamplesToEncode(const Audio::SamplesBuffer& samplesToEncode, quint8 channelIndex, bool isFirstPart, bool isLastPart){
+    //qCDebug(jtNinjamCore) << "Adding samples to encode";
+    QMutexLocker locker(&mutex);
+    chunksToEncode.append(new EncodingChunk(samplesToEncode, channelIndex, isFirstPart, isLastPart));
+    //this method is called by Qt main thread (the producer thread).
+    hasAvailableChunksToEncode.wakeAll();//wakeup the encoding thread (consumer thread)
+}
+
+void EncodingWorker::stop(){
+    if(!stopRequested){
+        //QMutexLocker locker(&mutex);
+        stopRequested = true;
+        hasAvailableChunksToEncode.wakeAll();
+        qCDebug(jtNinjamCore) << "Stopping Encoding Thread";
     }
+}
 
-    ~EncodingThread(){
-        stop();
-    }
-
-    void addSamplesToEncode(const Audio::SamplesBuffer& samplesToEncode, quint8 channelIndex, bool isFirstPart, bool isLastPart){
-        //qCDebug(jtNinjamCore) << "Adding samples to encode";
-        QMutexLocker locker(&mutex);
-        chunksToEncode.append(new EncodingChunk(samplesToEncode, channelIndex, isFirstPart, isLastPart));
-        //this method is called by Qt main thread (the producer thread).
-        hasAvailableChunksToEncode.wakeAll();//wakeup the encoding thread (consumer thread)
-
-    }
-
-    void stop(){
-        if(!stopRequested){
-            //QMutexLocker locker(&mutex);
-            stopRequested = true;
-            hasAvailableChunksToEncode.wakeAll();
-            qCDebug(jtNinjamCore) << "Stopping Encoding Thread";
+void EncodingWorker::run(){
+    while(!stopRequested){
+        mutex.lock();
+        if(chunksToEncode.isEmpty()){
+            hasAvailableChunksToEncode.wait(&mutex);
         }
-    }
-
-protected:
-    void run(){
-        while(!stopRequested){
-            mutex.lock();
-            if(chunksToEncode.isEmpty()){
-                hasAvailableChunksToEncode.wait(&mutex);
-            }
-            if(stopRequested){
-                mutex.unlock();
-                break;
-            }
-            EncodingChunk* chunk = chunksToEncode.first();
-            chunksToEncode.removeFirst();
+        if(stopRequested){
             mutex.unlock();
-			if (chunk){
-                QByteArray encodedBytes( controller->encode(chunk->buffer, chunk->channelIndex));
-                if (chunk->lastPart){
-                    encodedBytes.append( controller->encodeLastPartOfInterval(chunk->channelIndex));
-                }
-
-                if(!encodedBytes.isEmpty()){
-                    emit controller->encodedAudioAvailableToSend(encodedBytes, chunk->channelIndex, chunk->firstPart, chunk->lastPart);
-                }
-				delete chunk;
-			}
-
+            break;
         }
-        qCDebug(jtNinjamCore) << "Encoding thread stopped!";
+        EncodingChunk* chunk = chunksToEncode.first();
+        chunksToEncode.removeFirst();
+        mutex.unlock();
+        if (chunk){
+            QByteArray encodedBytes( controller->encode(chunk->buffer, chunk->channelIndex));
+            if (chunk->lastPart){
+                encodedBytes.append( controller->encodeLastPartOfInterval(chunk->channelIndex));
+            }
+
+            if(!encodedBytes.isEmpty()){
+                emit controller->encodedAudioAvailableToSend(encodedBytes, chunk->channelIndex, chunk->firstPart, chunk->lastPart);
+            }
+            delete chunk;
+        }
     }
-
-private:
-    class EncodingChunk{
-    public:
-        EncodingChunk(const Audio::SamplesBuffer& buffer, quint8 channelIndex, bool firstPart, bool lastPart)
-            :buffer(buffer), channelIndex(channelIndex), firstPart(firstPart), lastPart(lastPart ) {
-
-        }
-
-        Audio::SamplesBuffer buffer;
-        quint8 channelIndex;
-        bool firstPart;
-        bool lastPart;
-    };
-
-    QList<EncodingChunk*> chunksToEncode;
-    QMutex mutex;
-    volatile bool stopRequested;
-    NinjamController* controller;
-    QWaitCondition hasAvailableChunksToEncode;
-
-};
+    qCDebug(jtNinjamCore) << "Encoding thread stopped!";
+}
 
 //+++++++++++++++++ Nested classes to handle schedulable events ++++++++++++++++
 
@@ -187,6 +155,7 @@ NinjamController::NinjamController(Controller::MainController* mainController)
     mutex(QMutex::Recursive),
     encodersMutex(QMutex::Recursive),
     encodingThread(nullptr),
+    encodingWorker(nullptr),
     preparedForTransmit(false),
     waitingIntervals(0)//waiting for start transmit
 {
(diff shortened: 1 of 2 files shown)


elieserdejesus:
Nice work Tea!
Your comments/sections appears as checkboxes for me. What this means? It's a kind of readed/unreaded or agree/disagree? 

I have another idea to simplify this encoder stuff. I think the producer/consumer approach is too much. I'm thinking in try QtConcurrent::run. Every time we need samples to encode we just call a "encode function" using QtConcurrent::run. No more wait conditions, no more workers, threads, etc. Using this simple approach we can delete a lot of code. 
I can think in a negative effect: the chunk encoding order. QtConcurrent::run will create a thread (or use a thread from a pool) for each available audio buffer/chunk. So, is possible the second audio chunk is encoded and sended before the first chunk. This will mess the transmited audio stream. A possible solution is some kind of "serialization" in decoding. QtConcurrent::run return a QFuture, and I see a class QFutureSynchonizer in QtConcurrent namespace.

What you think about?

teamikl:
the checkbox is a task item to do. we can use the list in GitHub flavor markdown
https://github.com/blog/1375%0A-task-lists-in-gfm-issues-pulls-comments
asignee can see how much the work is progressed. when all are checked its time to close.
it's same as milestones.


QtConcurrent is higher layer approach, Yes code will be much simple. I agree it better choice too.
about saving the order of chunks, maybe QtConcurrent::mapped? or reduce has an order option.


This PR can be rejected, since the warning is small matter, I think no trouble for users.
or can be minimum patch to just remove the warning. since this topic is about warning,
not multi-thread encoding. I think main topics is about the destructor of the thread.
with QtConcurrent approach, no more use of thread directly, so warning also will be removed that time.

As the side of WIP PR practice, we see the merit, we can discuss about this early stage.
I am not doing those listed tasks yet, so I can save times, no waste works.
and also could shared those thread information.

teamikl:
- CHECKED: I am not testing the side effects by this changes.
  tested standalone version on windows
- CHECKED:  Is it safe to remove the destructor?
 with qtconcurrent approach, no need this
- CHECKED: review codes, small matters
  with qtconcurrent approach, worker class will be dropped. so not for now/here.

teamikl:
> It's a kind of readed/unreaded or agree/disagree?

the check status in official blog was explained (incomplete or complete)
However, it up to project how it's used.
I did strike && checked for the task is incomplete but the topic gone.

A rejected task also can be checked as complete, because not task in the thread anymore.
no need to count as task to show the progression.

using "verb" as prefix to explain the status, may be useful tip:

e.g task for @elieserdejesus (you can check when finished to read the url)
- [ ] read https://blog.qt.io/blog/2010/06/17/youre-doing-it-wrong/
- [ ] make new branch for concurrent-audio-encoding

About the coding style and unit test, please check when you read it.
they are off topics, no need to discuss on this thread.

elieserdejesus:
Tea, you ask **when Jamtaba increment the middle version?**
At moment the middle number is used to handle client/server incompatibilies. For example, if real time rooms are added we need change the server, and old clients will be incompatible. When adding something like real time rooms we change from 2.0.* to 2.1.*

teamikl:
Oh, rest of my comment was disappear, because of forget to escape &lt;&gt;

@elieserdejesus 
You can document that idea on wiki. because it's project local rule no one knows expect you.
(or you can just note the link, if other project documented that versioning scheme)

following known documented versioning rules can save the time, this is an easy scheme.
a lot projects on github use versioning &lt;major&gt;.&lt;minor&gt;.&lt;patch&gt;

in that case
Usually, such incompatible change is a big task, can be major version up.
Normal version up is minor version, small fix after release are up patch version.
API/protocol also can have own version, if need.

more info: [Semantic Versioning](http://semver.org/)

For JamTaba 1 to 2 was big changes, and you may think its 2.&lt;major&gt;.&lt;minor&gt; is similar.
This is not a request to changing your versioning. my focus is on sharing the management rules.

teamikl:
Now another topic, about the blocking of main thread (event loop)
This is a proof just possibility, I did not test it happen in a real case.

https://gist.github.com/teamikl/d384b14d5e91a6a80d33

- click "invoke method", "call direct" will works without lock
- check lock
  - click "invoke method", print invoke method OK and show "done" when unlocked
  - click "call direct" will freeze UI, can't unlock because of event loop was in blocking.
    (please kill process from task manager)

addSamplesToEncode method is called directly in main thread.
so that will block event loop when mutex was locked.
call via invokeMethod will aboid the blocking, patch here:


```diff

diff --git a/Client/src/NinjamController.cpp b/Client/src/NinjamController.cpp
index 48536c4..f79de03 100644
--- a/Client/src/NinjamController.cpp
+++ b/Client/src/NinjamController.cpp
@@ -242,7 +242,11 @@ void NinjamController::process(const Audio::SamplesBuffer &in, Audio::SamplesBuf
                             mainController->mixGroupedInputs(groupIndex, inputMixBuffer);
 
                             //encoding is running in another thread to avoid slow down the audio thread
-                            encodingWorker->addSamplesToEncode( inputMixBuffer, groupIndex, isFirstPart, isLastPart);
+                            QMetaObject::invokeMethod(encodingWorker, "addSamplesToEncode", Qt::DirectConnection,
+                                                      Q_ARG(const Audio::SamplesBuffer, inputMixBuffer),
+                                                      Q_ARG(quint8, groupIndex),
+                                                      Q_ARG(bool, isFirstPart),
+                                                      Q_ARG(bool, isLastPart));
                         }
                     }
                 }
```

teamikl:
CHECKED: thread context
now added the commit and proof code as a test

Ezeer:
seem to be very good , i am not aware of threads management in jamtaba actually , but what you have
found really interest me .
That change can work as is , without other files modifications ?
How can i use the patch please ? (in my fork  )

teamikl:
@Ezeer 

This patch was for my branch, i renamed s/encodingThread/encodingWorker/
so the invokeMethod snippet works just rename the variable name.

However, if you was working on branch, no need to add this patch to your branch.
or you will meet conflicts when merge to master.
(I do not know about how your "fork" schedules merge , not "branch" ?)

This thread was a practice of WIP RP (work in progress pull request),
I think this is a low priority task unless the thread's work become too complex.
current JamTaba code is working fine.

If you are interested in Qt5's thread management, two key points in this PR are

- [QMetaObject#invokeMethod](http://doc.qt.io/qt-5/qmetaobject.html#invokeMethod)
- [QThread#moveToThread](http://doc.qt.io/qt-5/qobject.html#moveToThread)

I suggest to read the article https://blog.qt.io/blog/2010/06/17/youre-doing-it-wrong/

@elieserdejesus shown his plan with QtConcurrent,
(ideally) code would be more simple with that higher layer API.
so This PR can be close without merge these commits.

Ezeer:
ok , i'll wait .
And read your links when i will have time to .
i already read " you're doing it wrong " .
thx tea .
^^

teamikl:
CHECKED: milestone request
no need if QtConcurrent approach would be chosen.
 
Now all tasks are checked. It's soon a time to close.

teamikl:
@elieserdejesus 
I removed "[WIP]" prefix from the subject. I had finished all my tasks in this PR.
So, now you can decide this PR merge or reject to close.
(a reject also a one of choice, the checkbox does not mean accept/agree, it's case by case)

If accept then assignee can "Merge pull request"
If reject then "close and comment" explain the reason as comment.
for this case now, mention about QtConcurrent approach as the reason.


As I am contributor, seem I have the permission to do that on the github system.
but I will not do it myself, This is an Assignee 's task. need their review and clearance.

teamikl:
*noticed the typo now*

teamikl:
EDIT: commit and patch
the 3rd argument for QMetaObject::invokeMethod should be
s/Qt::DirectConnection/Qt::QueuedConnection/

elieserdejesus:
Merged Tea, thanks!

teamikl:
@elieserdejesus 
Can you withdraw this merge ?
on current master this had a problem that I did not notice in my branch debug,
Perhaps the parent of thread or worker object, but for quick solution, drop this patch.

teamikl:
Sorry to use this closed PR thread.
This is a quick patch, I will make new PR if you need.

```diff
diff --git a/Client/src/NinjamController.cpp b/Client/src/NinjamController.cpp
index ef1aba9..d31f7c3 100644
--- a/Client/src/NinjamController.cpp
+++ b/Client/src/NinjamController.cpp
@@ -291,7 +291,6 @@ void NinjamController::stop(bool emitDisconnectedingSignal){
         encodingWorker->stop();
         encodingThread->wait();//wait the encoding thread to finish
         delete encodingWorker;
-        delete encodingThread;
         encodingWorker = nullptr;
         encodingThread = nullptr;
     }
@@ -352,8 +351,8 @@ void NinjamController::start(const Ninjam::Server& server, QMap<int, bool> chann
     processScheduledChanges();
 
     if(!running){
-        encodingThread = new QThread(this);
-        encodingWorker = new EncodingWorker(this, this);
+        encodingWorker = new EncodingWorker(this);
+        encodingThread = new QThread(encodingWorker);
         encodingWorker->moveToThread(encodingThread);
         QObject::connect(encodingThread, SIGNAL(started()), encodingWorker, SLOT(run()));
         encodingThread->start();
```

teamikl:
Sorry to use this closed issue thread.
this was not so simple, the problem was a life span of `inputMixBuffer` which located on a stack.
This is a quick patch for revert and fix.

```diff
diff --git a/Client/src/NinjamController.cpp b/Client/src/NinjamController.cpp
index ef1aba9..519281d 100644
--- a/Client/src/NinjamController.cpp
+++ b/Client/src/NinjamController.cpp
@@ -242,11 +242,7 @@ void NinjamController::process(const Audio::SamplesBuffer &in, Audio::SamplesBuf
                             mainController->mixGroupedInputs(groupIndex, inputMixBuffer);
 
                             //encoding is running in another thread to avoid slow down the audio thread
-                            QMetaObject::invokeMethod(encodingWorker, "addSamplesToEncode", Qt::QueuedConnection,
-                                                      Q_ARG(const Audio::SamplesBuffer, inputMixBuffer),
-                                                      Q_ARG(quint8, groupIndex),
-                                                      Q_ARG(bool, isFirstPart),
-                                                      Q_ARG(bool, isLastPart));
+                            encodingWorker->addSamplesToEncode(inputMixBuffer, groupIndex, isFirstPart, isLastPart);
                         }
                     }
                 }
@@ -291,7 +287,6 @@ void NinjamController::stop(bool emitDisconnectedingSignal){
         encodingWorker->stop();
         encodingThread->wait();//wait the encoding thread to finish
         delete encodingWorker;
-        delete encodingThread;
         encodingWorker = nullptr;
         encodingThread = nullptr;
     }
@@ -352,8 +347,8 @@ void NinjamController::start(const Ninjam::Server& server, QMap<int, bool> chann
     processScheduledChanges();
 
     if(!running){
-        encodingThread = new QThread(this);
-        encodingWorker = new EncodingWorker(this, this);
+        encodingWorker = new EncodingWorker(this);
+        encodingThread = new QThread(encodingWorker);
         encodingWorker->moveToThread(encodingThread);
         QObject::connect(encodingThread, SIGNAL(started()), encodingWorker, SLOT(run()));
         encodingThread->start();

```

teamikl:
the patch was not enough.
There was a side effect bug when close ninjam room.
the code did not terminate worker cleanly ?

elieserdejesus:
Tea, I see these problems here too. I change the worker parent to nullptr, change the QMetaObject::invoke call. These changes solve the more simple problems, but the bug when closing a room appears. 
I don't look at this problem (close room bug) yet, but I think the main thread is blocked by a lock. Another possibility is the call for **waitToFinish**. Maybe the worker is not finishing correctly, possibly running a endless loop.

teamikl:
the quick and simple solution is revert those merge commits.
we does not need to debug this deeply now if that will change to qtconcurrent in future.
at least, the previous code was working. and also revert master to deploy-able is high priority task.
(or can not make new branch for other work)


`git revert` will help the task

```sh
$ git revert -m 1 ad923e2
```

`ad923e2` is the merged commit.
after the revert, confirm the changes
```sh
$ git diff HEAD^
```

we should do this review and test process before merge. This is an Quality assurance topic though,
It's my mistake I was missing to mention about assignee/tester/reviewer 's roles on Pull Request.

There are several extra stuffs:
Nightly build, now tester has to do build those branch from source
  if there was auto build binary, a lot testers can join the test.
CI (continuous integration) can avoid this situation these human had to test manually.

teamikl:
no needed those commands, GitHub show "Revert" button

PR #197, opened 2015-12-11
Revert "Fix QWaitCondition warning by separate thread"

teamikl:
Reverts elieserdejesus/JamTaba#189

The patch had problems.
Since the task was low priority, does not need it now.
no debug, just revert to clean master.

### DIFF
--- a/Client/src/NinjamController.cpp
+++ b/Client/src/NinjamController.cpp
@@ -21,62 +21,94 @@
 
 #include "audio/samplesbufferrecorder.h"
 #include "Utils.h"
+#include <QWaitCondition>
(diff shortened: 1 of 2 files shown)


teamikl:
Ah so qucik!
I was going to write tasks before merge, now.

- [ ] build debug mode
- [ ] test enter ninjam room and leave the room

teamikl:
It's ok I will test these now.

elieserdejesus:
Tested here, working.

teamikl:
To test the encoding thread working, had to wait 3 intervals. It ok same here.

PR #204, opened 2015-12-17
Sanitize cache data

teamikl:
This is a continue from 0b608fe4417ffa50fdbe954baa46c0bb05014edc

- could be binary format?
 - if there is a chance to change the format, let discuss it first.
   QDataStream may support easy serializing QMap collection. (if its ok to read whole data)
 - [x] test QMap serialization
 - [x] versioning / signature header
- where sanitize
 - [x] CacheEntry setters
 - [x] read
 - [x] write (incoming user name from ninjam)
- format/validation
 - [x] escape format characters ";" and `endl`
 - [x] user name, research what characters are allowed
 - [x] check ip address format (maybe with mask)
 - [x] float / unsigned int 8bits / bool
- [x] error action
 - what should do if invalid data came.
   I think now is report the error, skip and use default value.
- [x] how to test (where to put test code)
  - [x] unit test
  - [x] functional test

Please comment if there were missing tasks. This is the first stage, need more details.

### DIFF
--- a/Client/src/gui/NinjamRoomWindow.cpp
+++ b/Client/src/gui/NinjamRoomWindow.cpp
@@ -295,7 +295,7 @@ void NinjamRoomWindow::on_channelAdded(Ninjam::User user, Ninjam::UserChannel ch
 
     //remembering user track controls (gain, mute, pan, boost)
     UsersDataCache* cache = mainController->getUsersDataCache();
-    CacheEntry cacheEntry = cache->getUserCacheEntry(user.getIp(), user.getName(), (int)channel.getIndex());//a empty/default cacheEntry is returned if the user is not cached
+    CacheEntry cacheEntry = cache->getUserCacheEntry(user.getIp(), user.getName(), static_cast<quint8>(channel.getIndex()));//a empty/default cacheEntry is returned if the user is not cached
 
     if(!trackGroups.contains(user.getFullName())){//first channel from this user?
         QString channelName = channel.getName();
--- a/Client/src/persistence/UsersDataCache.cpp
+++ b/Client/src/persistence/UsersDataCache.cpp
@@ -122,7 +122,7 @@ UsersDataCache::~UsersDataCache(){
     }
 }
 
-CacheEntry UsersDataCache::getUserCacheEntry(QString userIp, QString userName, quint8 channelID){
+CacheEntry UsersDataCache::getUserCacheEntry(const QString& userIp, const QString& userName, quint8 channelID){
     QString userUniqueKey = getUserUniqueKey(userIp, userName, channelID);
     if(cacheEntries.contains(userUniqueKey)){
         return cacheEntries[userUniqueKey];
--- a/Client/src/persistence/UsersDataCache.cpp
+++ b/Client/src/persistence/UsersDataCache.cpp
@@ -49,9 +49,65 @@ QDataStream & operator>>(QDataStream &stream, CacheEntry &entry)
 
 
 //+++++++++++++++++++++++++++++++++++++++
-CacheEntry::CacheEntry(QString userIp, QString userName, quint8 channelID, bool muted, float gain, float pan, float boost)
-    :userIp(userIp), userName(userName), channelID(channelID), muted(muted), gain(gain), pan(pan), boost(boost){
 
+CacheEntry::CacheEntry(const QString& userIp, const QString& userName, quint8 channelID)
+{
+    setUserIP(userIp);
+    setUserName(userName);
+    setChannelID(channelID);
+    setMuted(DEFAULT_MUTE);
+    setGain(DEFAULT_GAIN);
+    setPan(DEFAULT_PAN);
+    setBoost(DEFAULT_BOOST);
+}
+
+void CacheEntry::setUserIP(const QString& userIp)
+{
+    if (ipPattern.exactMatch(userIp)) {
+        this->userIp = userIp;
+    }
+    else {
+        qCDebug(jtCache) << "invalid ip address: " << userIp;
+    }
+}
+
+void CacheEntry::setUserName(const QString& userName)
+{
+    if (namePattern.exactMatch(userName)) {
+        this->userName = userName;
+    }
+    else {
+        qCDebug(jtCache) << "invalid name: " << userName;
+    }
+}
+
+void CacheEntry::setChannelID(quint8 channelID)
+{
+    this->channelID = channelID;
+}
+
+void CacheEntry::setMuted(bool muted)
+{
+    this->muted = muted;
+}
+
+void CacheEntry::setPan(float pan)
+{
+    this->pan = qBound(PAN_MIN, pan, PAN_MAX);
+
+    if (pan != this->pan) {
+        qCDebug(jtCache) << "pan " << pan << " was out of range";
+    }
+}
+
+void CacheEntry::setBoost(float boost)
+{
+    this->boost = boost;
+}
+
+void CacheEntry::setGain(float gain)
+{
+    this->gain = gain;
 }
 
 UsersDataCache::UsersDataCache()
--- a/Client/src/persistence/UsersDataCache.cpp
+++ b/Client/src/persistence/UsersDataCache.cpp
@@ -147,6 +147,7 @@ void UsersDataCache::loadCacheEntriesFromFile(){
     QFile cacheFile(cacheDir.absoluteFilePath(CACHE_FILE_NAME));
     if(cacheFile.open(QFile::ReadOnly)){
         QDataStream stream(&cacheFile);
+        stream.skipRawData(12); // skip header
         stream >> cacheEntries;
     }
     qCDebug(jtCache) << "Tracks cache items loaded from file: " << cacheEntries.size();
@@ -160,7 +161,15 @@ void UsersDataCache::writeCacheEntriesToFile(){
     if(cacheFile.open(QFile::WriteOnly)){
         qCDebug(jtCache) << "Tracks cache file opened to write.";
         QDataStream stream(&cacheFile);
+
+        // TODO: move thise to header. this is a qucik dirty implementation
+        quint32 signature = 0x4a544232; // "JTB2"
+        quint32 revision = 1;
+        quint32 offset = 12; // sizeof(signature) + sizeof(revision) + sizeof(offset)
+        stream << signature << revision << offset;
+
         stream << cacheEntries;
+
         qCDebug(jtCache) << cacheEntries.size() << " items stored in tracks cache file!";
     }
     else{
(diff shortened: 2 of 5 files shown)


elieserdejesus:
**serialization and binary format**
The idea of QMap serialization and binary format is very good. It's ok read/write whole data. It's just one read at constructor and one write at destructor. We need change the CacheEntry class to be serializable too, right?

**sanitizing**
My vote is to sanitizing in read.

**Uni tests**
Man, this is up to you. As you can see in this project "I love unit tests" :)  What you propose to work with tests? Some "C++Unit" very cool framework? I never check the QtCreator test features, maybe we can use some test frawework well integrated with QtCreator.

teamikl:
>  We need change the CacheEntry class to be serializable too, right?
current CacheEntry class can stay.
it's overloading QDataStream operator's "<<" and ">>"
I am going to follow this manual way
http://doc.qt.io/qt-5/qdatastream.html#reading-and-writing-raw-binary-data

## sanitizing

Ok, "write" mean here is the user name which came from Ninjam.
BTW, the cache is for ninjam only (now it's) ? or assume real-time jam in future too ?

mixing them is a possible but that may make a limitation by the dependency.
for example, can not use ";" in name is an unnecessary limitation for real-time session protocol.

## validation / range check

- user ip \d+\.\d+\.\d+\.(\d|x)   "x" for mask
- user name [a-zA-Z0-9\_] another characters allowed ?
- channel ID  range? or max limit (greater or equal than zero?)
- muted    (0|1)
- gain/pan/boost float ... what their range?

## Unit tests

I need research about the QtCreator integration. I did not use the IDE.

Qt has own test library that support simple test and benchmarks.
supports keyboard/mouse simulation for widget test too,
so I believe it's the best Qt friendly frameworks.
but not sure QtCreator support it or not. I will search plugins.

elieserdejesus:
**the cache is for ninjam only (now it's) ?**
It's for realtime (future) too. 

**can not use ";" in name is an unnecessary limitation for real-time session protocol**
I think this is not a problem. Allow users names as [a-zA-Z0-9_] is enough for ninjam or realtime.

**validations**
- channel ID range: [0, ?]. Zero is included, but I'm not shure about the max limit. 32 is enough?
- gain and boost are 32 bits floats. 
- pan: at moment pan is handled as 9 integer values [-4, 4], zero is the center, but the type is float. If I remember correctly the type is float because when I start Jamtaba the pan widget was a horizontal slider. At moment the pan widget is a QDial truncated to only 9 values. But I think make sense keep the type as float, because we can change the QDial to allow more values without change the rest of the code.

teamikl:
Ok, thanks for those details.

## about limit of channel ID

If it was the index number of channel in ninjam remote channels
then the limit is uint8_t in ninjam protocol (0 to 255).
but rare to use so much (maybe used in session mode)
before reach the limitation, I think ninjam client had a problem when too many channels.
test with 32 channels is enough.

elieserdejesus:
Tea, you are correct, the channelID in CacheEntry class is the index number of remote channel. In ninjam server codes I'm using the Qt **quint8** to handle the 8 bits value. In other places I'm using **int**, but for nothing, just lazzyness. The channel index value is always a 8 bits integer (unsigned).

teamikl:
tested QMap serialization
https://gist.github.com/teamikl/eb8033a27f560e018ff7

teamikl:
## Research Qt testing frameworks

- http://doc.qt.io/qtcreator/creator-autotest.html
    commercial version has the plugin

- http://dragly.org/2013/04/19/setting-up-unittest-with-qt-creator/
    I read this about project file/directory structure and another test frameworks
    GoogleTest can be next choice, there was 3rd party plugin for QtCreator integration
    (EDIT: the plugin v3.6.0 did not work with QtCreactor v3.4.2)

- https://bugreports.qt.io/browse/QTBUG-23067
  one of missing feature of QtTest
  QtTest does not have macro to support multiple test suites in a single projects.
  to run multiple tests, have to write main function manually.

QtTestLib is light weight, missing several stuffs other frameworks has.
e.g. utility to test private member in non-friend class etc.
However, the advantage is no need extra installation, I will start with QtTestLib.

teamikl:
## cache file format's version and migration

I am thinking the header for binary format. signature and version.
Usually, binary format should have a version info for migration in future expansion.

- I will ignore read current tracks_cache.bin (line based semi-colon separated format)
  - because, there is no simple way to detect the format.
  - please announce delete previous cache file manually, or uninstaller can do the task?
- use application version  or  file format's own version
  the file format should have own version, because application version may cause some troubles.
  - no need to update the version if the format did not updated.
  - patch release which may not increment version number.
    as I see current JamTaba versioning, there is no patch version?
    It's like JamTaba2.MAJOR.MINOR

migration strategy can be simple, the logic I assume now. (NOTE: case without `break`)

```cpp
switch (version) {
case 1:
  // do 1 to 2 conversion
case 2:
  // do 2 to 3 conversion
case 3:
  // latest version do nothing
  break;
default:
  // error unknown version
  break;
}
```

elieserdejesus:
**please announce delete previous cache file manually**
The "track remember" feature is not released yet, so we can change or delete the tracks_cache.bin without problems for users.

elieserdejesus:
Tea, I see just two unchecked items in your list. Please tell me when you are done and thanks for the good work.

About inline: I will refactoring these inline stuff some day. This is Interesting, today (with modern compilers) inline is useless, good to know.

teamikl:
I see, current working state is
now I have tested this branch, had another problem. (freeze on enter ninjam room)
done the module part almost, but not integrated test yet. I need to add more task.

about read/write
can skip those tasks.
now validate in setter methods can check both cases.

If you need schedule to merge this PR, please set milestone.

teamikl:
there was a rest topic, about version of binary file format.
## binary file header

it's better to add this in early stage.

- signature
- version

there is parameter "byte order" often used in binary format.
now use QDataStream default.  (big endian)


### why need version

is for a future expansion.
e.g. we may want to remove outdated or non used records.
non-global ip-address and name are not unique, they may change some time.
think a plan to add several fields to the record, for statistics.

- count of how many times used.
- created date
- last looked up

the previous implementation is a delimiter separated values, could add more fields into the tail of  line.
for binary format, if there had version information, it's easy to migrate.

teamikl:
CHECKED: read/write -- setter methods checks both
ADDED: functional test, now trouble in integration. (when enter to room)
ADDED: binary file format header (header struct and versioning)

teamikl:
> About inline: I will refactoring these inline stuff some day.

I think no need to remove them for Just refactoring, it's just hints for compilers.
-  omit `inline`, write method body in declaration is inline by default (without inline keyword)
- remove `inline` keyword when overloading the method
- separate .h/.cpp file and if you want setter methods to mark inline explicitly, can use `inline` keyword.

NOTE: There was another way to solve the situation, (moved setter implementation to cpp)
I removed the inline keyword in declaration, because the method did not have `inline`
but could add `inline` to the method implementation.

elieserdejesus:
Tea, the milestone is setted.

teamikl:
I did not make sandbox for testing. I will clean up data dir under AppData, and try it again.

if there was no problem then, rest task I will commit is add header to the binary.
(simple one, signature and version)

teamikl:
> the milestone is setted.

Ok, there was no date set on milestones, tho.
this was the last task? I can work this weekend.

elieserdejesus:
Yes Tea, this is the last task for v2.0.10. I'm not setting dates for the milestones, I think dates are too much "business" oriented :)

teamikl:
Done functional tests manually.

How do you think about versioning of the file format?
If you want make v2.0.10 release soon, that can be a future topic.

for now I will add this header for future backward compatibility (testing now)
[signature 4byte][revision 4byte][offset 4byte]

elieserdejesus:
Tea, I have no experience with this "file format" question. I have no ideas, and I have some questions:
[signature 4byte] - what is the signature? Some "jamtaba magic word" to identify a Jamtaba config file?
[revision 4byte] - this is ok, it's simple to understand.
[offset 4byte] - ????

teamikl:
Yes, I will explain it.
for now the contents are not so important.
this should be documented on wiki or source/header files

[signature 4byte]
  - the magic number is "JTB2" (hex dump the ascii numbers)
  - i looked somewhere use the short word. added "2" is for 4byte alignment.
  - uint32 read/write as uint32 is make possible endian
    - NOTE: QDataStream use big endian by default.
[revision 4byte]
  - ok, increment when the format updated.
[offset 4byte]
 - for future expansion too. its for the header. when header size was changed
   seek to the offset position, and start read contents

teamikl:
added a header to the binary.
so please remove previous tracks_cache.bin file. (or try to read the invalid file, what happen?)
it's not support the backward compatibility during debug work-in-progress revisions.

### about versioning of file format
It's an optional topic, can be a future task.

### unit test
missing several tests, no tests for the header.

teamikl:
## check list

- [ ] functional test: load/save remote channels settings
- [ ] error behavior: when invalid data came, report value and skip inside setter methods.
- [ ] test specs: please check those default values and validate the range. I am not sure the specs.
  - no worry about type check now, QDataStream will read data correctly.

elieserdejesus:
**so please remove previous tracks_cache.bin file. (or try to read the invalid file, what happen?)**
I see this problem here, Jamtaba crash when starting if the QDataStream try read the old tracks_cache.bin (text based, not binary).  But no worries, nobody has this file at moment, it's not released yet.

**Funcional tests**
I tested your last PR version (without the file header) and works. The test was simple, just change the gain, pan and mute, leave the server and enter again. All stuff was working.

elieserdejesus:
Tea, the *constexpr* don't compile in my msvc 2013. I tested in mingw and works, but fail in msvc 2013. You can see the details [here](http://stackoverflow.com/questions/20264644/constexpr-not-compiling-in-vc2013). 
Please change the **constexpr** to a more common const declaration (const int DEFAULT_VALUE). Off couse I can install the msvc 2013 patch to solve this problem, but this is a pain and not worth it.

teamikl:
Ok, I will change

- [ ] `constexpr` to `static const`
- [ ] correct namespace (static member of Persistence::CacheEntry)
- [ ] separate declarations and data body

elieserdejesus:
Tea, I tested the code and working nice here. I will merge here and just change the name UsersDataCacheHeader to CacheEntryHeader. Thanks for your nice work! Job done!

teamikl:
I had one more commit within this PR.
sorry for this unscheduled task, let me do it now it's a quick topic.

```cpp

if (signature != UsersDataCacheHeader::SIGNATURE) {
    qDebug() << "invalid users data cache file";
} else {
    switch (revision) {
    case 1:
        if (size == UsersDataCacheHeader::SIZE)
            stream >> cacheEntries;
        break;
    default:
        qDebug() << "unknown users data cache revision";
    }
}
```

the two qDebug() are debug information, users will not see in release build.
does users need to know if the cache file was broken? or just skip reading?

teamikl:
Ah, thank you fix rename task on your side.
I was in progress. It's ok this snippet will be a note, use in future change.

PR #218, opened 2016-01-03
[WIP] Refactoring ninjam code

elieserdejesus:
This code (ninjam protocol related) was the first code I work in Jamtaba. In that moment I was re-learning C++ and learning Qt. Now I'm revising the code and is very fun rewrite these stuff and apply other technics, change some design, etc.

The task list is big, but fun:

- [x] Renaming 'ugly' names
- [x] Use const references in parameters
- [x] Review 'stack versus heap'
- [x] Replace STL code with Qt similar (just to avoid mix)
- [x] Use the override keyword to replace the bad "virtual"s in overrided functions.
- [x] Clean the old school style casts
- [x] Remove throws
- [x] Remove static stuff
- [x] Add tests


elieserdejesus:
Using rebase -i to squash some related commits.

elieserdejesus:
Using rebase -iagain to reduce the number of commits. After a lot of squashed commits and tons of conflicts merge the number of commits are small enough :)

PR #240, opened 2016-01-18
Cleaning up project to attempt Linux builds

abramhindle:
Linux tends to use case sensitive file systems. The project is rife with case insensitive spellings. These commits attempt to address that problem. Mostly renamings.


elieserdejesus:
Hi @abramhindle , thanks for your contribution, I will integrate soon.

PR #257, opened 2016-02-04
Linux port

elieserdejesus:


elieserdejesus:
@abramhindle , if you are interested Jamtaba is compiling in linux now. I'm testing in Ubunt 15 (inside Virtual Box).

abramhindle:
Cool. I'll give it a try :)

PR #282, opened 2016-04-06
Adding Portuguese translation (pt_BR)

elieserdejesus:
- Translating to portuguese

- Unchecking the "translatable" checkbox in some forms properties like 'WindowTitle' when the propertie is not used.

- Fixing some missing parameters in translated strings.

- Remove tr(...) from non translatable strings. For example, I'm considering "JamTaba" a non translatable string.


elieserdejesus:
@teamikl and @jonjamcam , when translating for Portuguese I see many little problems for future translators.
Some strings are marked as "translatable" in Qt forms because this is the default option, but sometimes these strings are just "Window Title", and this title is never showed in JamTaba GUI. I just remove these strings from "translatable" list.

I see some parameters missing in tr(...) codes. These small issues are fixed in this PR, but I think you both need update your translation files, because some strings are removed and others are changed a little bit.

From this point we have a more mature GUI from the translation point of view and these changes will be minor in future versions.

elieserdejesus:
this PR is integrated!

jonjamcam:
I had to use lupdate to update the .ts files. Now the changes appear in Qtlinguist

teamikl:
@elieserdejesus 

Can u mark translatable all user visible text?
also the initial text which would not show in application,
but debug preview in linguist it's used.

notr



I am also updating .ts file now, I saw type="vanished" attribute
they could be removed by lupdate -no-obsolete

elieserdejesus:
@teamikl , the **server name** label in the picture need be translatable? Make sense? These things (names) are translated in Japanese? Or **ninbot.com** will be **ninbot.com** in everywhere? I have no idea about how these things works in another countries.

About the other label (the "5 musicians" text). This label make sense as translatable, but I have no idea about how do feedback to translators. If we check the label as translatable translators will see a "5 musicians" string in Qt linguist, but this string is just a placeholder, and the string format is very different from the format used in run time (picture translated in Portuguese):
image

Any idea about how to have "real" feedback about this piece of translation?

ps: if I understand correct **lupdate -no-obsolete** will remove the old strings, right?

teamikl:
@elieserdejesus 

The initial text can follow the real format like: `tr("e.g. 5/7 musicians 120 BPM 12 BPI")`
it's not necessary to translate, but the placeholder is useful that show in preview,
it's information for translators.

some of literals which need to save the identity

"Jamtaba" as no translatable, its ok.

about "DLL (\*.dll)"  the "(\*.dll)" part is used as literal in program
`tr("DLL") + QStringLiteral("(*.dll)")`

key sequence for shortcut etc, they also can be literal (non translatable text).

about "b" and "#"
it's translatable text, I would like them translate to music sign.
There are Japanese translated terms for sharp and flat also
I will not use the term though Because English sharp/flat is more popular,
but I do not know this for other countries situation.

For Japanese, English musical/audio terms are used
but they are shown as Katakana letters (one of Japanese letters set for foreign words) often,
so need them translatable.

we need discussion for all the text/situations?
I think was a simple solution here, mark as translatable and
if translators think English term is ok then just they can keep the same text.

about -no-obsolete options
I did not check about old strings,
it look like the old strings are marked as type="vanish" when lupdate without -no-obsolete options.

elieserdejesus:
@teamikl , I agree with 
`I think was a simple solution here, mark as translatable and
if translators think the English text is ok then just they can keep the original text.`

teamikl:
> "Jamtaba" as no translatable, its ok.

I got another idea now.
use qApp->applicationName() can remove the hard coded string literal.

Issue #283, opened 2016-04-06
Option to select language (other than default)


jonjamcam:
Now that the internationalization is working it would be good to have the option to change to other available language if the user chooses to (like the translate function is able to).

elieserdejesus:
Done in PR #284

jonjamcam:
Language Menu working fine! (to be able to compile I had to use some old .qm files for now)

elieserdejesus:
Jon, if you run the **lrelease** command the .qm files will be generated. Can you try?

jonjamcam:
Yes! qm files generated

PR #280, opened 2016-04-06
Spanish translation files

jonjamcam:

PR #284, opened 2016-04-06
Language selection

elieserdejesus:
- [x] Create a menu to select the language
- [x] Retranslate the UI when a new language is selected reacting to Qt **changeEvent**. 
- [x] Remember the preferred language
- [x] Change the auto translate language in chat when user select another language

Solving the issue #283.


elieserdejesus:
Just to save the reference, [This text](http://www.informit.com/articles/article.aspx?p=1405555&seqNum=3) is interesting, specially the last part:

`For widgets and dialogs developed with Qt Designer, the uic tool automatically generates a function similar to our retranslateUi() function that is automatically called in response to LanguageChange events.`

So, it's not necessary reimplement the **changeEvent** for Qt Designer forms, good news!

teamikl:
I forget about install translator for VST Plugin version.
the language select can solve it?
likely move the install translation code in standalone main() to some where common app starting event.

jonjamcam:
@elieserdejesus @teamikl I was reviewing https://github.com/elieserdejesus/JamTaba/pull/284/commits/8843d57cc4283deb3518177e3fa4e8ffbab1cd3f and it looks very similar to what happened here https://github.com/elieserdejesus/JamTaba/pull/280 . The Problem was that all translation files were changed. The solution was to revert the changes of the mistakenly changed files with: https://github.com/elieserdejesus/JamTaba/pull/280#issuecomment-206110541 but need to be expanded to the other trnslation files as well. Am I on the right track?

teamikl:
@elieserdejesus 
I felt we need to improve the management of TS files resource.
those update *.ts/*.qm file log had became a noise for preview. (a lot diff lines)

- drop *.qm from repo, it is generatable by lrelease command
  - NOTE: then developer have to invoke the command manually or add the task to .pro file
    this also back to which .pro should have TRANSLATIONS again, though.
- use lupdate -locations none to eliminate all line numbers in the TS files
  - [Best practice for Qt translations and git](http://stackoverflow.com/questions/28220472/best-practice-for-qt-translations-and-git)

just idea for now. any ideas to reduce the noise in commit log.

@jonjamcam 
no need to run lupdate on this branch topic, on your local environment.
just use existing TS/QM files.

elieserdejesus:
@teamikl , about **I forget about install translator for VST Plugin version.**. In this PR the translator installing is moved to **MainWindow** class, this class is used in VstPlugin too, so this will work (non tested yet :)).

I agree about the noise with .qm files. I will remove the files and add a entry in .gitignore.

elieserdejesus:
@teamikl , I read your link about Best practices for Qt translations in Git, but the solutions are more "problems" in my opinion. The idea is generate a temp .ts file, remove the line numbers and push this temp. But you need remember to push the "no lines" version.  Unfortunally I can't see my brain remembering this :) Someday a file with a lot of lines changed will be pushed.

I have no ideas now, time to sleep, but we can discuss some idea to improving the translation workflow.

teamikl:
> Someday a file with a lot of lines changed will be pushed.

It's not a critical problem.

Ideally, code review stage should work to block that happen. committer can revert.
Technically solution, git can have hook to check it before commit. but it a bit advanced use.

Research real projects, e.g. qtcreator has perl utility script to check TS file.
but some of TS files contains filename/lineno, some does not have.

NOTE: their translations path was ./share/qtcreator/translations, like a POSIX directory layout.

@jonjamcam 
INFO: ca0293b dropped qm files
now we need manually to run `lrelease ./PROJECTS/Jamtaba.pro` to generate qm files before build.

we can add this as pre-build task, but it's still on discussion stage for now.

elieserdejesus:
@teamikl , If I'm understanding correctly we need two different solutions (we have 2 little problems):
1 - how to generate the **.qm** on the fly and avoid put these files in repo? 
2 - How we can avoid mess the translation files (.TS)?

For the first question the **pre-build task** sounds nice for me.

For the 2nd question I'm thinking here (after a coffee): 
    Imagine I'm adding a new translatable string in the code. I will call **lupdate** to update my TS file, but this will update other TS file too. For me this make sense, because a new string is added in the source, and the other TS files need some revision, right?
    The other problem I can think now is the scenario where the source is changed and the **context lines** in TS files are obsolete. In this scanerio I have 2 questions:
    1 - If I run **lupdate** for all TS files your translations will be messed? All strings need be translated again? I really hope this is not the case ... :)
    2 - If I update just my TS file (your translations are untouched) your TS files are using wrong **context line numbers**, right? Your .qm files (generated by problematic TS files) will work?

I'm putting these question here to generate some discussion and better understanding of the scenarios.

elieserdejesus:
@teamikl and @jonjamcam , I finished the dynamic translation tasks. But after integrate the last changes in Jon translation this branch show some conflicts. The conflicts are in TS and qm file. I just ignore the problems in .qm file and drop the file. The conflicts in TS file are more complicated, because I see tons of line changes and some "vanished" translations. I solve these conflicts trying to use the last changes in Jon TS file, but I'm not shure if it's all ok.

I think use **lupdate -locations none** before push will be the best solution. This will remove the line number and will avoid a lot of changes in every translation.

jonjamcam:
@elieserdejesus If you need, just revert my last PRs. I have saved my translation files and I can submit them again. no problem.

elieserdejesus:
@jonjamcam , this PR is integrated in master. I change all .TS file to remove the line numbers using the command **lupdate -locations none** in **Jamtaba.pro**. Please try open your translation file in Qt Linguist and check if your work is still valid (I hope yes).

elieserdejesus:
I pushed the TS files without the line numbers in master. Tested these TS files in my Qt Linguist and works. So, I think we don't need these line numbers at all.

jonjamcam:
I checked es_CL in QTlinguist and I got some old entries like:

old audio

then I did lupdate -locations none Jamtaba.pro.

but old entries still appear.

Now I have a trick to be able to update only one .ts file. First:

rem translationspri2

then use  lupdate -no-obsolete. All "#" files are skipped and now old entries are gone!

translationspri3

elieserdejesus:
Very good idea Jon. I think the old entries are my fault when solving the last conflicts. Possibly I pick the wrong lines when merging the files.

jonjamcam:
So TODO is revision of the translations for missing parts. I updated my es_CL locale, but the .qm file was not included.

no problem to generate .qm files locally?

elieserdejesus:
no problem Jon, the idea is generate the .qm files locally to avoid put these binary files in the repository. 

At moment (until we have an automated way) we need run the command **lrelease** before compile.

teamikl:
@elieserdejesus 
> If I'm understanding correctly we need two different solutions (we have 2 little problems):

Yes

> For the first question the pre-build task sounds nice for me.

this has the problem, which .pro file should have the task.
./translations/translations.pri can contain the task. but which .pro should include it.

1. lupdate Jamtaba.pro
it's not pre-build when open the subproject Standalone.pro

2. each Standalone and VstPlugin pros 
conflict when Jamtaba.pro ... 
management cost of the duplicated code
e.g if add another options for lupdate, need same edit twice

3. make .pri for Common or resources
then include in Standalone.pro/VstPlusing.pro
if it's possible, I think include guard like if qm file were new then not run lupdate.

> For the 2nd question I'm thinking here (after a coffee): 
Imagine I'm adding a new translatable string in the code. I will call lupdate to update my TS file, but this will update other TS file too. For me this make sense, because a new string is added in the source, and the other TS files need some revision, right?

I am not testing this yet, hope the `-location none` options to be the solution.

> The other problem I can think now is the scenario where the source is changed and the context lines in TS files are obsolete. In this scanerio I have 2 questions:
> .. snip ..

Yes, wrong context line number, when source code are update.

then need to lupdate to update TS files. I think with `-location none`
if the context was changed (file layout changes etc, refactoring may have)
eventually TS files also will have the changes.
it's an acceptable situation TS file updated and info commit.

but it's can be optional, 
If lupdate on pre-build task, then the old revision TS file will be updated at build time.

There is another question who commit TS file update? and when?

there is not good scenario if committers have to care the TS files update every times.
I think that make decrease development speed.

anyway **update TS file as separate commit, that can avoid the noise in review.**
or we can delay the TS files update until release coordinate time.

teamikl:
the side effect of `-locations none`
no source code preview on Linguist.

so if translation editors want to use preview,
we need to share this workflow

1. lupdate -no-obsolete   # this will generate context details, -no-obsolete is an optional.
2. linguist
3. lupdate -locations none   # remove filename and line number from context
4. commit

elieserdejesus:
@teamikl and @jonjamcam , about the discussions:

`this has the problem, which .pro file should have the task.`
 , at moment we have a .pri to common and resource, and this .pri is included in VstPlugin.pro and Standalone.pro. I think this commom.pri is the best place to put pre-build stuff.

`There is another question who commit TS file update? and when?`
I think **when** is more simple. I think the best moment to update translation files is **before** release a new version. The translation tasks always will be the last tasks in the stack. So we can change everything, and when the things are in the right place we work in the translations. Maybe we don't need work in translation files in every release.

about `who commit TS file update`
I think the author(s) of translation file? I can't see another option here. I'm missing something?

`anyway update TS file as separate commit, that can avoid the noise in review.`
Sounds perfect for me. I have bad habits in git, sometimes my commits are very mixed. I'm trying to improve.

`or we can delay the TS files update until release coordinate time.`
Exactly, work in translation only at the end of task list.

`the side effect of -locations none`
Ahhh, the lines are usefull for something :)

`we need to share this workflow

lupdate -locations relative # generate line numbers
linguist
lupdate -locations none # remove line numbers from context
commit`

This work sounds nice for me, but I think the first step (lupdate -locations relative) can be optional. For example, in general I'm not interested to see the source code context because I know these contexts. I'm thinking in Magnus and DoubleBass (I will invite these guys to help in translations for German and French), and these guys are not coders, so I think the source code context is useless to them.

PR #289, opened 2016-04-07
pt_BR and en fix 

jonjamcam:
@elieserdejesus I did and update to the pt_BR locale that had some "vanished" entries and this fix too the "outpus" mistake. The es locale just sets the language to en. Check it out. It can save you some time.


elieserdejesus:
Merged in master, thanks Jon!

Issue #291, opened 2016-04-08
English only (no translation available) items


jonjamcam:
I found this two items that are still fixed in English only.

![clicks and license](https://cloud.githubusercontent.com/assets/15310433/14372511/b2b3fd7c-fd17-11e5-92fb-b7db077015d0.jpg)

elieserdejesus:
Jon, just to clarify:
1 - your idea is translate the text items in custom metronome combobox?
2 - The licence button has no text, just an icon. But this button has a tool tip text. Your idea is translate the tool tip or translate the licence text?

jonjamcam:
1.- the click names text (i.e. finger snap -> spanish, jap, etc)
2.- The idea is to let non-english speakers read the license text in their native language.

teamikl:
@jonjamcam 

the license text is NINJAM server side resource.
private server may return their custom text, that can not translate as client side resource.

about Creative Commons license, there are translations.
link to the page is helpful? I think people does not read the license every time.

the license text use case in Japanese private server,
we use describe server local rules there, like README for new comers.

teamikl:
about link to CC license page
If creative commons licence URL was detected then link to the language page with locale options

e.g. http://creativecommons.org/licenses/by-nc-sa/2.5/deed.ja

NOTE: http://creativecommons.org/licenses/by-nc-sa/2.5/deed.ja_JP was redirected to "deed.ja"

jonjamcam:
@teamikl I think your solution is good enough. This is as you say not read too often. And most (if not all) servers are some kind of CC license.

elieserdejesus:
good ideas CC links!

elieserdejesus:
@teamikl , another piece missing translation is the country names in the public servers/rooms list or inside a room. I try some approaches with QLocale and the **nativeCountryName**, but no success. You know some trick in this field? I think translate the country names is very important, some english country names are very non intuitive in comparison with native language.

teamikl:
@elieserdejesus 

I think QLocale does not support such translation for out of resources.
Qt's resource ./Src/qttranslations/translations/*.ts does not have translated country names.

Perhaps, there is a translated resource library ?

elieserdejesus:
@teamikl , I found a way, a free webservice:

![image](https://cloud.githubusercontent.com/assets/1012741/14404006/8263f6aa-fe42-11e5-9371-ccafb6fb6af9.png)

[...]

PR #293, opened 2016-04-08
Implement CC license link idea

teamikl:
- [x] ~~get locale string, system locale or user selected language~~
- [x] ~~click link will call shell exec on windows, need review for security~~
- [x] CC license version 3.0 and 4.0 links worked with the same url format. 

implement idea on #291.


teamikl:
about security of QMessageBox

QMessageBox the default [textFormat](http://doc.qt.io/qt-5/qmessagebox.html#textFormat-prop) is [Qt::AutoText](http://doc.qt.io/qt-5/qt.html#TextFormat-enum)
if setText detect html tag then set format Qt::RichText implicitly then
click a hyperlink with Qt::RichText will call shell execute on windows platform.

current JamTaba code also may have this.
showing NINJAM license text case in JamTaba:
- only server admin can set the license text (unless the server has security hole)
- require user's click action to invoke

other effects?
- [ ] other platforms openUrl implementation
- [ ] other use case: QMessageBox, or everywhere remote strings shown in QLabel widgets

here proof code:
https://gist.github.com/teamikl/65e870322c60567fb1371dcaa93b9135

teamikl:
> get locale string, system locale or user selected language

This was not a possible scenario.
user selected languages are only available if translation resource exists.
so use getTranslation info here was useless, others will see English which not equal to their locale.

> click link will call shell exec on windows, need review for security

I added escape code to avoid this (like `s/</&lt;/')
so I will not research details about this, I check this task as consumed
with modify it strike styles (canceled task).

teamikl:
@elieserdejesus 
all scheduled tasks has completed. now it's stage for test/review, please check.

jonjamcam:
@teamikl Just tested. The link is working, but the language chosen is the one of my system (es) and not the one Jamtaba is using. I think this is good enough. In my opinion no need to change it.

elieserdejesus:
@teamikl and @jonjamcam , tested here and is working. The link point to the portuguese licence translation here.

elieserdejesus:
@teamikl , can I integrate this PR in master? I check the commits and see no issues to review, it's nice for me.

teamikl:
@elieserdejesus 
Yes, if review/tested was passed. please!

elieserdejesus:
It's integrated in master, thanks!

PR #297, opened 2016-04-09
Fix server disconnect by special characters

### LINKED ISSUE #294: Server disconnects if some special characters are used in the insrtument box
I tested this with some accented letters like Á, ü, á. Looks like the server rejects this kind of characters if used in Jamtaba. This is important in relation with the internationalization feature because new users that use almost any other language than except English will be tempted to use this letters (or others) and will get a disconnection.

I tested Reaninjan on this issue and there's no problem of disconnection there:

image

When this type of characters are used in reaninjam, they appear in Jamtaba wrongly decoded, but test with two instances of reaninjam  is necessary to check.

NOTE: Jamtava VST 2.0.13 used to test

teamikl:
The problem was reported by @jonjamcam at #294

The details was payload size gap in NINJAM protocol message,
payload size was counted QString::size() but real data size was utf-8's size.

this commit will not fix the same issue for userName.
userName in NINJAM protocol is ASCII only.

fix #294

### DIFF
--- a/src/Common/ninjam/ClientMessages.cpp
+++ b/src/Common/ninjam/ClientMessages.cpp
@@ -98,7 +98,7 @@ ClientSetChannel::ClientSetChannel(const QStringList &channels)
     payload = 2;
     channelNames.append(channels);
     for (int i = 0; i < channelNames.size(); i++) {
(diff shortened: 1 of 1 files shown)


elieserdejesus:
Thanks Tea!

Issue #296, opened 2016-04-09
Allow user name change


elieserdejesus:
In the issue #295 we start a discussion about improvements in user name change. At moment Jamtaba users can't change the user name without restart Jamtaba.

My initial idea is:
1 - Change the user name widget (a simple label at moment) to an editable/input widget (QLineEdit). Using this approach users can type your names in this field.
2 - First time Jamtaba is started there's no user name saved in .json file. So we can show a string "your name" in the current "your controls" field. 
3 - The user name dialog will be eliminated, and I will love this :)  If users don't change "your name" the user name in the server will be "your name". It's a very good name, like "anon" :)
4 - If the user name is changed when jamming (connected in a server) we need show a message like "Your new user name will be used when you connect in a new server". It's not possible change the user name in Ninjam protocol.

@jonjamcam and @teamikl , please share your thinkings!

jonjamcam:
I like it! My only concern is that is you allow changing user name TEXT while connected to a server and not perform the actual change, this can confuse users about what name they are actually using (specially with the VST Plugin as no other feedback is available). A more intuitive approach I think is:

1.- Do not allow the name to be changed if connected to a server, so if someone tries to change the name during a session, then a message will warn him this can't be done while connected. Maybe use a  gray or white color for the name as a visual feedback for this.

2.- when not connected, use green color for the name (or the background) so users know they are allowed to edit it (to be used in the next session).

jonjamcam:
> If users don't change "your name" the user name in the server will be "your name"

May I suggest (if possible) to use the computer name to generate a correct identity automatically (similar to what git console does for commits). I know this is a windows feature, but not sure about mac or linux

elieserdejesus:
`use the computer name to generate a correct identity automatically`
Nice idea, I'm 99% shure this is possible and using Qt this will be easy to do.

**I have another idea for the feedback for name changing:**
When users are jamming we can disable the user name input field and show a lock icon in the field
https://cdn1.iconfinder.com/data/icons/free-98-icons/32/lock-16.png

teamikl:
> 1. Do not allow the name to be changed if connected to a server
:+1: not change during a session.

my suggest, read-only and editable attribute for the text input.

read-only and enable/disable are a bit different stuff.
I am not sure Qt's stylesheet support those attributes.

> It would be very strange to enter a server full of "your name" players in the future! :D.

it's same there were a lot "anon" in the past!

[...]

PR #298, opened 2016-04-09
[WIP] Allow user name change

elieserdejesus:
- [x] when users type blank spaces replace with "_"

Solving the issue #296


elieserdejesus:
@teamikl  and @jonjamcam , I work in the user name change issue. The result is:

1 - Showing the placeholder text when user name is not stored in .json file:
image

2 - When connecting in a server the user name dialog appear only if user name is empty. The user name in the dialog is **guessed** from user path in the machine. In my example is exactly my common user name, a perfect "guess" :D
image

3 - Locking the user name field and showing the connected user name:
image

jonjamcam:
looks great!

jonjamcam:
@elieserdejesus Does this new username selection method filter invalid names? It may be a concern that not every character is accepted by the protocol as @teamikl said (here https://github.com/elieserdejesus/JamTaba/issues/296#issuecomment-207897560):

>about non-ASCII alphabets for username, currently that is NINJAM's limitation.
>JamTaba can implement however have to communicate with existing server and another clients.

teamikl:
I arrived this issue,

## tested ff2d9dc7201e898b51577a2324c5ac8768ddb5ec

that validation worked.
ReaNINJAM allow another characters but it was replaced to "_" server side.

## review MainWindow::getGuessedUserName()

- the function does not require gui, can be utility ?
  - make easy to write the unit test without the gui dependency
- can use QDir::dirName()
- empty is returned in rare case, e.g. root user and set custom home "/".
  - how about environment variable "USER" ? it's more direct info.

jonjamcam:
I tested this branch. No problems! Just one thing. It looks the max character for the user name is truncated by the server to 16 characters:

image

But the input box in Jamtaba allows much more than that (same as reaninjam).

@elieserdejesus , @teamikl  Need to set a limit or leave it just like that?

teamikl:
@jonjamcam 

Yes, for now (NINJAM only) it's useful users can notice that has limitation before connect.

There is another perspective, the max length limitation was NINJAM server's.
Real time room (in future) does not need to follow the limitation.

Implementation info for @elieserdejesus 
QLineEdit has maxLength property, or regex in validator can have the limit by {2,16}

NOTE: the max length will count QString::size, it's not equals to bytes in multi bytes chars.
this is not problem since it's ASCII only, but if the server assume that 16 bytes then 
limit by regex validator can check the byte size.

Even not set the limit, set that property to no limit explicitly is better code.
When contributor developers in future read the code,
they can know from code it's not just forget to set the limit.

the default max length for QLineEdit is 32767 (max of signed int16)

@elieserdejesus 

This is github topic, I see many piece of small commits
do you consider to use this new feature? (or already use?)
https://github.com/blog/2141-squash-your-commits

I am interested how it's shown on github.

elieserdejesus:
@teamikl , about
`ReaNINJAM` allow another characters but it was replaced to "_" server side.`
Exactly, the blank spaces are replaced in ninjam server. But now I'm thinking is better remove the blank space from jamtaba to avoid different names in ninjam server and Jamtaba.

**About MainWindow::getGuessedUserName() reviews:**
`the function does not require gui, can be utility ?`
Shure @teamikl. A new utility class or a namespaced function? I see many C++ references suggesting namespaced functions. 

`can use QDir::dirName()`
To remove the "indexOf" stuff, right? Good idea.

`empty is returned in rare case, e.g. root user and set custom home "/"`
`how about environment variable "USER" ? it's more direct info`
the USER variable sounds good for me. We can do this in a crossplataform /Qt way?

@jonjamcam says
`user name is truncated by the server to 16 characters. Need to set a limit or leave it just like that?`
I think is better limit the user name field to 16, even for real time rooms this lenght is enough. @teamikl, **setMaxLenght** or change the **regex**? Or both?

`This is github topic, I see many piece of small commits do you consider to use this new feature? (or already use?)`
I see this feature Tea, now the github merge button is showing a combobox with 2 options: merge and squash. I never use the squash option in git hub directly. I will squash some commits in this branch before merge in master.

elieserdejesus:
Using the environment variables works here:

`QString MainWindow::getSuggestedUserName() const
{
    //try get user name in home path. If is empty try the environment variables.

    QString homeDirName = QDir::home().dirName();
    if (!homeDirName.isEmpty())
        return homeDirName;

    QString userName = qgetenv("USER"); // for MAc or Linux
    if (userName.isEmpty())
        userName = qgetenv("USERNAME"); // for windows

    if (!userName.isEmpty())
        return userName;

    return "";
}`

teamikl:
@elieserdejesus 

> setMaxLenght or change the regex? Or both?

can use both, I mention here is setMaxLength(16) is incorrect for this use,
that does not mean 16bytes for multi bytes chars (although its ASCII only)

> A new utility class or a namespaced function? 

can be static function, separate from MainWindow
the point is if make unit test for the function, it require gui module.
but the function can be Independence.

unlike Java, no need to use class as namespace.
if no need to make the instance.

> //return a better stuff here?

it's rare case to arrive there and safe to be empty,
ninjam server allow it and gives name anon

I thought the squash option is same as squash in rebase.

push --force to remote should be avoid as much as possible,
because the commit hash changes, pages in github may be unreachable.
I expect the squash merge does on remote.

teamikl:
>  safe to be empty,

I forgot the function was for suggestion, it not decide name. and there was validator.
I have no idea. or just anon  default name for ninjam server.

if it was different then make two default names.
JamTaba's default name and NINJAM's default name.

@elieserdejesus 

about length limit in regex

can be {1,16} or {,16} ? I know a case user name "b" is used, 
and empty also acceptable.

elieserdejesus:
@teamikl  and @jonjamcam , the discussions/reviews are implemented in the last commit in this PR branch. I decide to use {1, 16} in the regex. Empty names are not allowed. I think the "anon" culture is a mistake in ninjam history, I think it's better not encourage anonymous. People can type anything as a "fake" name, why use empty/anon names?

I set the maxLenght to 16, I think is enough at moment because the user name input is ASCII only. I added a comment warning another developers about the 16 characters issue if non ASCII user names are allowed. I think is enough at moment.  

Can I integrate these commits in master or I miss some review?

jonjamcam:
@elieserdejesus 

Tested. All Working Solid!

teamikl:
@elieserdejesus 

> QString MainController::getSuggestedUserName

about the name space
I wanted to point is when write the unit test, the test code requires the controller instance.
the function can be static function in somewhere utility namespace.

> about setMaxLength 16

I worried is they may mislead length == bytes though,
this is ok with the line comment :+1:

> "anon" culture is a mistake in ninjam history

Actually, I had explain to new comers every time. they think someone name "anon"

{1,16}  so JamTaba does not allow blank default name?
it's good require users input name, if they does not want fixed name, they can type "anon" too.

> return "";  //return a better stuff here?

I had worried scenario JamTaba/Ninjam both use different default names.
that make user confuse, return empty string and force user to input something is ok.

I did not test this case yet, the zero length is invalid user name.
will not start connect to server?

jonjamcam:
> I did not test this case yet, the zero length is invalid user name.
>will not start connect to server?

@teamikl I tried hard to force an empty name (in the "suggestion box" and in the name box), but that's not accepted. It reverts to "type your name here" or to the previous name if you already chose one.

elieserdejesus:
@teamikl 

> the function can be static function in somewhere utility namespace
The function is static in **MainController** class, no instance required at moment to unit testing. But maybe the MainController class require including a lot of Qt modules just to test. This class (MainController) is inside the **Controller** namespace. We can move the function to this namespace, outside the class. What you think?

> {1,16} so JamTaba does not allow blank default name?
it's good require users input name, if they does not want fixed name, they can type "anon" too.
Exactly Tea, no empty names allowed. If somebody hired me to write a "ninjam troll detector" my first heuristic will be the empty or anon name :D 

I think Jon solve the last comments about empty names, right?

teamikl:
@elieserdejesus 

yes, that unnecessary module dependency for the utility function.
This can be delay until when make unit test for getSuggestedUserName function.

> I think Jon solved the last comments about empty names, right?

It's ok if tested.

elieserdejesus:
@teamikl and @jonjamcam , I added the pencil icons and a tooltip in user name field (when read-only). This is the result:
image

jonjamcam:
Nice!

jonjamcam:
@elieserdejesus  Looks like the spacebar is nor working anymore for entering the name in the username box

elieserdejesus:
@jonjamcam , the blank space is not allowed. Ninjam servers not allow blank spaces, these characters are replaced with "_" as you know. I think is best not allow spaces in Jamtaba to avoid a name with spaces appearing in Jamtaba and another (with "_") in ninjam server.

jonjamcam:
Yes. I agree with that, but is there a possibility to be able to use the spacebar to enter the "_"?. Some people write their complete names and It's hard to get used to use the _ instead of the space.

elieserdejesus:
> but is there a possibility to be able to use the spacebar to enter the "_"?
Is possible @jonjamcam , this is a very good idea.

elieserdejesus:
@jonjamcam , the last commit in this branch is replacing the spacebar pressing with a '_' character. The result is very cool, thanks for the nice idea.

jonjamcam:
Nice! The best of both worlds! THX.

elieserdejesus:
PR merger in master.

teamikl:
@elieserdejesus 
nice work! I have 1 additional. can add tooltip for the input field?
that describe allowed letters and explains the space change to underscore.

> Solving the issue #296
did not close #296 automatically?

elieserdejesus:
@teamikl , the issue is not automatically closed. Unfortunally github does not bind issues and PRs. 

I will add a new issue to add the tool tip in user name fied.

teamikl:
I see, that was only commit message and specific format.

PR #301, opened 2016-04-11
[WIP] Translating and caching country names

elieserdejesus:
- [x] Use nekudo API to translate the country names
- [x] Notify the GUI when a user IP is resolved to a translated country name
- [x] Use separated files to cache country codes and translated country names
- [x] Create a "cache" dir to organize the cache files
- [x] If the Jamtaba2 folder in "appData/Local" is deleted a wrong country name cache file is created. This file name is missing the language, looks like some bug in the language selecion.
- [x] Move the tracks_cache.bin file to the new cache dir.
- [x] Check if the tracks_cache.bin file exists in old dir and copy the file to new 'cache' dir.
- [x] Refactor the Configurator class to remove the checking for "Standalone" or "VstPlugin".
- [x] Test the new cache dir in Vst Plugin.
- [x] Add version fields in country cache files.
- [x] Add the checking and tests proposed by Tea in the [comment](https://github.com/elieserdejesus/JamTaba/pull/301#issuecomment-209025570)


jonjamcam:
@elieserdejesus , @teamikl I was taking a look at my own file "cache.bin" I opened in notepad and it shows this:

image

I highlighted one example to make my point. You have IP;(country name in english);(country code). My idea is to use only the country code because the english country name is now just one of the possible options.

So now we have to translate just a country code, for example FR. For this we can use Tea's resource. We choose the language to be translate to, for example portuguese. So now we only need to look FR inside the pt.json file:

image

and there it is! Isn't this the simplest way? and all the resources are already at our disposal!

teamikl:
@elieserdejesus 

isn't it too much documentation?  (anti-pattern)
I think summary is enough,
we can discuss approach on here github issue.

- detail on beginning stage may be worst work if the approach was rejected.
- for non English speaker, simple code is more easy to understand.
  - if document just follow code, then 
- maintenance cost for the long document
  - when code is changed, then document also must be updated

that document was look like describe specs in BDD (behavior driven development)

about review the approach itself, I will take a time. i go to sleep now, gnite!

elieserdejesus:
@teamikl 
> isn't it too much documentation? (anti-pattern)
I think yes, in general I never write these stuff :D

@jonjamcam, I'm shure the nekudo api is the simplest approach at moment because we don't need make two http requests (1 for nekudo and another for the github json) and wait for the 2 responses.

about the advantages you listed:

> Smaller local cache file size
Are you thinking in cache only the country code, right? If this is the case all the country name translation need be "on the fly" using http requests, and no cache for country names.

> Instantaneous translation (not need to wait 20 secs for the nekudo service doing the translation every time the language is changed).
Nekudo is fast (for a free service), it's about 2 or 3 seconds at maximum to return the translations. The slow update is in Jamtaba code. The country names are refreshed every 30 seconds when a fresh versions of the public rooms list is received. When the layout change (change the view mode, or maximizing) the country names are updated too, and if the country names are cached the result is Instantaneous. The best solution I can see is change the current aproach to a event driven approach. Every time we received a translation the GUI is **notified** and the country name is translated. No need to wait until the next refresh cicle. But this solution is very "fat fingers", a lot of files will be changed.

jonjamcam:
Ok. I'm convinced. When you say fat fingers I tremble in fear! :D. I'm glad the solution is not that hard to code.

teamikl:
@elieserdejesus 
> 1 for nekudo and another for the github json

not http request, I assumed there was the json as embed resource.
but nekudo could have the translated name, then no need the json.

there are several trade off stuff between embed or external service.
at this moment, I think it's enough if safe for fallback case when service not unavailable.

@jonjamcam 

> Many users already have this files in their computers, so maybe it's not a good idea to change this files.

Solution for this, the name "cache.bin" is missing the subject what kind of cache.
we can drop "cache.bin" and use new file name.

@elieserdejesus 
I suggest for the binary file format, add signature and version info for top of file.
that will help migration task in future.

Do you plan Qt version update to 5.6 ? the new update had version number class.
maybe big task ? the update require re-build qt for static and libraries etc...

teamikl:
about drop cache.bin, there was possible trouble case
it's for debug build previous version. cos the app config directory is shared.
no sandbox path for debug.

elieserdejesus:
@teamikl and @jonjamcam , please check this branch. Now the country names are translated almost in real time when language is changed. I added the tasks in the PR description.

I do a rebase and removed the old first commit in this branch, please drop your old branch and recreate to pull these fresh commits.

jonjamcam:
Tested. Working OK. It's really fast! 

+1

jonjamcam:
here : https://commons.wikimedia.org/wiki/File:Pencil_BLACK-01.svg

elieserdejesus:
@jonjamcam 

> maybe delay in request? or server refresh?

Delay in request. We are using the free service in nekudo, possibly sucessive requests from the same IP are delayed. Sometimes I experienced some delays too.

jonjamcam:
Overall this is a great improvement. Not just the speed, but once the country is cached it works for all ip on the same country! Very efficient use of resources!

teamikl:
@elieserdejesus 

tested, the working was ok  [[captured animation]](https://cloud.githubusercontent.com/assets/146129/14469620/8911656a-0120-11e6-8596-cfbd9a6d26d9.gif)
I have code review, that for unusual cases.

- more strict sanitize check.
  - the passed value will be show in HTML, then should not allow '<' '"' etc 
  - and unit test for the proof the sanitize check is working.
- around json parse, missing return value check.
  - e.g. if non-json data was returned, if the structure was changed,  
     at least should avoid null pointer exception.

elieserdejesus:
Ok @teamikl , I added one more task in the PR description to do these checkings and tests. Thanks for your review!

teamikl:
@elieserdejesus 
Ok, just minor comment fix.

> in the last PR comment.

the "last" pointer will move if topic grow and I added new, you can get the link to comment from date.
https://github.com/elieserdejesus/JamTaba/pull/301#issuecomment-209025570

elieserdejesus:
@teamikl , just a git little question: at moment this branch is conflicting with master. How you solve this problem?

My approach is: checkout the conflicting branch and **git rebase master**, solve the conflicts (I'm using [Meld](http://meldmerge.org/)), and now the problem: **push -f** to replace the old and conflicting commits in this branch. Off course this works, but as you know I'm rewriting a public git history.If no body is using the branch we have no problem, but sometimes this is not the case.

What you think about?

teamikl:
@elieserdejesus 
your question is about "rebase/push -f" ? or solve merge conflict?
or solve the conflict by rebase or marge?

### merge conflict
- `git merge master` on translating-country-names branch
- fix the conflict and commit

will solve the conflict, if you don't want the merge commit log then rebase though ...

### rebase and push -f

are you sure it's push -f for translating-country-names branch then
I think the effects is limited. no one make branch from the branch.
Although, if we had code review on commit comment,
they also will gone. but we did not use this moment.

If you think more strict rule "do not push -f to remote" then it's late.
there was a practice rebase before push commits to remote. it's safe case to use rebase.


this is a trade off stuff, if you think rebase then check the golden rules of rebasing
https://www.atlassian.com/git/tutorials/merging-vs-rebasing

jonjamcam:
@elieserdejesus, @teamikl  I did something to test the conflicts of this PR and it worked. It looks now it has no conflicts anymore.

 Can you guys check please if what I did is OK? It was an experiment part of my learning process...

Summary:

1.- I coded line by line all changes form this PR into a branch called "translating-country-names-manual-fix" in my fork.

2.- I isolated this line that was having a conflict from the file (that's why it's //):

WebIpToLocationResolver.cpp
L92         //emit ipResolved(ip);

3.- I tried to find the cause of why this line didn't work in my merged fork, but it did work in the PR, so I found the problem was a difference in the file:

 IpToLocationResolver.h

this line was missing:

L51     Q_OBJECT

4.- Finally I added the line above and the PR worked.

Hope this help solve the conflict. Please test my branch and let me know.

https://github.com/jonjamcam/JamTaba/tree/translating-country-names-manual-fix

teamikl:
@jonjamcam 
I think solve the conflict was not the main topic, though.
It's good you learn git too. git provide more easy way

0. assume you have added elieserdejesus's as remote, and `git fetch (remote name)`
1. you can 'git check translating-country-names` would make local branch from the remote
   (if you git version was not old)
2. `git merge master` will do the merge automatically, and CONFLICT is reported
3. (this case manual fix) open the reported file in editor, the conflicted part is shown like this
```cpp:
<<<<<<< HEAD
signals:
    void ipResolved(const QString &ip);
=======
    static QString getSuggestedUserName();
>>>>>>> master
```
4. edit to solve the conflict by manual, then save
5. test build on this stage.
6. if local test passes then `git add`  then commit.

conflicts are solved, for these small change, it's easy task.
@elieserdejesus worry was another situation, there is another way to solve the merge.
on that case, if others working on the branch from same repo then solving conflict on that way
may have side effects for others.

btw, I did not see Q_OBJECT is missing.
we are looking different branch ?

jonjamcam:
well....that's how I'm (ab)used to learn :D. I just try to do it and see what happens. it's true that the Q_OBJECT is there, but for some reason when I do checkout Elieser's Branch I have a compile problem and the solution took me into that direction. Donno. Anyway I need to be more in touch with code if I want to learn. It's been a good ride so far....

elieserdejesus:
@teamikl , thansk for the discussion about git. I understand the side effects of rebase and **push -f**.
The **git merge master** possibility is new for me. I use **gitk** to vizualize the history and using **git merge master** I'm worried about a messed git history because the commits timestamps. I will test soon.

elieserdejesus:
@teamikl and @jonjamcam , the conflict is solved now in this branch. I used the **git merge master** proposed by Tea and works fine. The commits in this branch are behind master commits, but it's the expected behavior. Tea, thanks again for share your knowledge!
image

elieserdejesus:
@teamikl and @jonjamcam , my tasks in this PR are done. Some tests added, checking possible invalid json data, added revision number for country name and code cache files, cache files moved to "cache" dir, etc.

teamikl:
@elieserdejesus 
I read code now, revision number is good enough this use case.

about header size, is it necessary? for alignment?(i think no need for this case)
or do you plan reserve extend space for future header change?

binary format often use "offset" there, it's seek to 12 and start read contents.
this can save file format compatibility without change the program code side.
If another program use the file set the value 16 with some additional info,
if it was offset then read contents will still work. I know this is out scenario, though.

about cache's revision check
I think it's not so critical, that is always safe to remove cache implicitly.
(without notification for users, they does not need to care the cache)
If revision was unexpected then the cache file just can be reset?

teamikl:
about header size
it was not in the latest commit, forget it.

about revision check,
I have question about qCritical 's behavior, it exit program by environment.
which behavior is expected there?

about reset cache, that was ok. load empty entries and new cache is created.

teamikl:
I have local test now, json object contains key check is ok,
now null check, to see watch happen toString() for Null value object.

teamikl:
null value object toString() returned an empty string. it's expected ok.

I think if release build will not exit by qCritical, then it's now ok to merge.

jonjamcam:
I have some compile errors. Already did the clean, run qmake, rebuild steps. This is what I get (MinGW x86):

image

teamikl:
@jonjamcam 
thank for the test, I did not meet this then I had noticed I was checking old revision.
I think this is MinGW issue, MSVC passed build.

should be fixed `s/Location::sanitize/sanitize/`

jonjamcam:
:+1: working now.thx.

elieserdejesus:
@teamikl 

> I have question about qCritical 's behavior, it exit program by environment settings. which behavior is expected there?

**qFatal()** exit the program, qCritical() will add a critical message in the log. This can be usefull in future when users have some problems with caching.

> about reset cache, (if not exit) that was ok. load empty entries and new cache is created.

Exactly, no need to reset the cache, reset is the default behavior.

elieserdejesus:
@teamikl , I added a last commit. I forgot to check **size** and **signature** in CacheHeader.

teamikl:
@elieserdejesus 
what the size is for ? I though that code was dropped...maybe i read was wrong revision.
if header has the **size** field then I have comment https://github.com/elieserdejesus/JamTaba/pull/301#issuecomment-210274382

elieserdejesus:
@jonjamcam , the Mingw compilation issue is fixed now.

@teamikl 
> what the size is for ?

I just follow your approach in the class UsersDataCache. I decide keep the size to preserve your idea about alignment. At moment the size is useless, but we can use in future. I can imagine a class hierarchy for CacheHeader. If a subclass add more fields the size will be different from 12 bytes, and the subclass can override **CacheHeader::isValid** to check this new field. I think keep the size field is still a good idea.

teamikl:
@elieserdejesus 
I see that my code had the size too.
the case I think assumed the size use as **offset** like this:
[SIGNATURE][REVISION][SIZE][ ... size will allow insert data here... ][DATA]

good idea about the subclass.
> I think keep the size field is still a good idea
Agree, keep the size field and no use.

elieserdejesus:
A squashed version of this PR is merged in master, job done @teamikl and @jonjamcam , thanks!

PR #313, opened 2016-04-15
Built-in-click-names-correction

jonjamcam:
The idea is to make the names more universal and simple.


elieserdejesus:
It's integrated in master Jon.  I squash the two commits, you will see just one commit in the master.

Issue #276, opened 2016-04-01
GUI do not respond when launching Jamtaba in full screen mode and then exiting full screen mode


jonjamcam:
It works OK if you return to full screen mode again. So to test the bug:

1.- launch Jamtaba and set full screen mode
2.- close Jamtaba and launch again
3.- change to normal mode
4.- GUI not responsive.
5.- Go to full screen mode again
6.- GUI now OK

elieserdejesus:
@jonjamcam , I can't reproduce this bug here (using the JT 2.0.13). 
How you detected the non responsive GUI?

jonjamcam:
@elieserdejesus OK my bad explanation. 

So to test the bug:

1.- launch Jamtaba and set full screen mode
2.- close Jamtaba and launch again
3.- change to normal mode **USING F11** (if you use the menu option, the bug is not triggered)
4.- GUI not responsive.
5.- Go to full screen mode again
6.- GUI now OK

elieserdejesus:
Jon, I still can't reproduce the bug. These are my steps:
1 - Launching JT and set to full screen (tested using the menu and F11)
2 - closing and launching again
3 - pressing F11

no bug here, I can use the buttons, listen the rooms, etc.

jonjamcam:
@elieserdejesus Please try using ony F11 to change views:

1.- launch Jamtaba and set full screen mode **USING F11**
2.- close Jamtaba and launch again
3.- change to normal mode **USING F11** 
4.- GUI not responsive.
5.- Go to full screen mode again
6.- GUI now OK

elieserdejesus:
@jonjamcam and @teamikl , I really can't reproduce the bug. My steps with more details:

0 - I'm using the JT 2.0.13 to test, the official released version installed in my machine. I'm not running from Qt Creator.
1 - Launching JT **in full view** (not mini node) and setting to full screen (**F11**)
2 - closing (**using ESC**) and launching again
3 - backing to normal mode pressing **F11**

no bug, working fine.

jonjamcam:
@elieserdejesus  ok. I will try tomorrow to test in another laptop with win10.

teamikl:
not happen to me
I tested official release 2.0.13 win64 standalone only.

which version @jonjamcam tested ?
VST ? about they press key action I think is VST had keyboard hook code.

BTW, ESC quit program!?
I know this behavior was same in Ninjam standalone client, and I wanted this to improve.
[...]

PR #316, opened 2016-04-16
Fixing the bug when opening in fullscreen

elieserdejesus:
- The issue #276 report a bug related with fullscreen. While testing this bug I see another related bug: when close Jamtaba using full screen and reopen the fullscreen is cropped in a very buggy way.

  **The bug context is**: In MainWindow::initialize() the QTimer::singleShot trick is used to avoid freeze the gui when loading many vst plugins in initialization. Using the QTimer::singleShot the GUI is opened, and after 1 milisecond the plugins are loaded (gui is opened when plugins are loading). Without this trick with QTimer::singleShot the **GUI is not opened until the last plugin is loaded**.

 - **The real problem is**: run the window initialization using QTimer::singleShot was spliting the window initialization in two separated calls, and this calls was out of sync.

- **The solution**: now the MainWindow::initialize() is calling doWindowInitialization() using QTimer::singleSlot. The doWindowInitialization() functions  **virtual** and is overrided in MainWindowStandalone (inheriting from MainWindow). **Now the entire window initialization is done inside doWindowInitialization()**, and **no more** problems with **out of sync** function calls.

### DIFF
--- a/src/Common/gui/MainWindow.h
+++ b/src/Common/gui/MainWindow.h
@@ -34,7 +34,7 @@ public:
     MainWindow(Controller::MainController *mainController, QWidget *parent = 0);
     virtual ~MainWindow();
 
-    virtual void initialize(); // this is overrided in inherited classes
+    void initialize();
 
     void detachMainController();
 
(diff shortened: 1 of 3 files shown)


jonjamcam:
Working OK here. Now the fullscreen mode does not crop and the bug seems to be gone! Tested also deleting the jamtaba.json file and no problems.

teamikl:
Tested and works ok
for me current master also did not happen the bug. it was only v2.13.0

Issue #309, opened 2016-04-14
User name text is behind pencil icon


elieserdejesus:
![image](https://cloud.githubusercontent.com/assets/1012741/14544687/a5e6fc12-0270-11e6-8342-b49255e21602.png)

jonjamcam:
May I add a suggestion for the pencil? 

![image](https://cloud.githubusercontent.com/assets/15310433/14554268/31ca5890-02c2-11e6-9eac-844b0a034039.png)

Circle a bit wider and red

jonjamcam:
Now it looks that the first ten characters fit nicely into the box:

![image](https://cloud.githubusercontent.com/assets/15310433/14580625/9af205d8-03a9-11e6-8deb-dea2395e55ee.png)

, So what to do with the remaining six? Some ideas:

1.- Move the pencil up when the 11th character is typed:

![image](https://cloud.githubusercontent.com/assets/15310433/14580646/8d441934-03aa-11e6-8a63-733a3745622c.png)

Another issue here is that the 16 characters allowed do not fit entirely i the window, so may be decrease the size of the font combined with the moving up of the pencil can solve it.

elieserdejesus:
@jonjamcam and @teamikl , I reduce the pencil icon a little bit and align in top. This is the result:
![image](https://cloud.githubusercontent.com/assets/1012741/14583224/9ef444f0-03f1-11e6-87af-78d40d03a404.png)

In my opinion is not good enough. It's difficult put these elements in the right place because we have limited horizontal space. I think the better/simplear approach at moment is show the icon only when the user name is read-only. In this state is no problem if the user name text is behind the icon.

I have another idea: remove the "local tracks collapse button" to gain some horizontal space. Put the "collapse option" in View menu as a menu item. But some users are used to collapse using the button, we are removing a possible interesting thing from the GUI.

Another ideas?

jonjamcam:
How about this?

![image](https://cloud.githubusercontent.com/assets/15310433/14583294/1718d0ca-03f4-11e6-82f0-f0007462d890.png)

elieserdejesus:
I like the second idea, will fit very well in "mini mode", and this mode is usefull in VstPlugin.
The old string in top of local controls was "Your Controls", I think we can use this string again. When we need translate this string again :)

PR #317, opened 2016-04-17
Fix user name behind pencil

elieserdejesus:
Fixing the issue #309

@teamikl and @jonjamcam , I think the user name layout is good enough now. I one of the Jon ideas (the second) in #309.

Please check the brach when you have some time!

### DIFF
--- a/src/Common/gui/MainWindow.cpp
+++ b/src/Common/gui/MainWindow.cpp
@@ -136,7 +136,7 @@ void MainWindow::showPeakMetersOnlyInLocalControls(bool showPeakMetersOnly)
     ui.localControlsCollapseButton->setChecked(showPeakMetersOnly);
     updateLocalInputChannelsGeometry();
 
-    ui.userNameLineEdit->setVisible(!showPeakMetersOnly);
+    ui.userNamePanel->setVisible(!showPeakMetersOnly);
     ui.labelYourControls->setVisible(!showPeakMetersOnly);
 }
 
@@ -685,8 +685,14 @@ void MainWindow::enterInRoom(const Login::RoomInfo &roomInfo)
 void MainWindow::setUserNameReadOnlyStatus(bool readOnly)
 {
     ui.userNameLineEdit->setReadOnly(readOnly);
+
+    QString icon = !readOnly ? "pencil.png" : "pencil-read-only.png";
+    ui.labelUserNameIcon->setPixmap(QPixmap(":/images/" + icon));
+
     QString toolTip = readOnly ? tr("Your name cannot be edited while jamming!") : tr("Only Letters, numbers, hyphen and underscore allowed! Spaces will be replaced by an underscore.");
     ui.userNameLineEdit->setToolTip(toolTip);
+    ui.labelUserNameIcon->setToolTip(toolTip);
+
     if (readOnly) {
         if (ui.userNameLineEdit->hasFocus())
             ui.userNameLineEdit->clearFocus();
--- a/src/Common/gui/MainWindow.cpp
+++ b/src/Common/gui/MainWindow.cpp
@@ -137,6 +137,7 @@ void MainWindow::showPeakMetersOnlyInLocalControls(bool showPeakMetersOnly)
     updateLocalInputChannelsGeometry();
 
     ui.userNameLineEdit->setVisible(!showPeakMetersOnly);
+    ui.labelYourControls->setVisible(!showPeakMetersOnly);
 }
 
 void MainWindow::updateLocalInputChannelsGeometry()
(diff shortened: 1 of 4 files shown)


jonjamcam:
Tested. It looks the upper text "Your controls" is not centered:

image

image

elieserdejesus:
@jonjamcam, It's centered now.

teamikl:
> QString icon = !readOnly ? "pencil.png" : "pencil-read-only.png";

I think this is a safe case, because on the same expression. (not tested)
There is compiler specific issues around questionmark operator with QString implicitly conversion.
https://wiki.qt.io/Coding_Conventions#Compiler.2FPlatform_specific_issues
use QStringLiteral (or QLatin1String) explicitly always will be good practice.

Tested Minimam case (View Mode -> Mini View does effect to the name input field width?)
the width is up to font and alphabet. how about set something monospaced font ?
user-name-width

teamikl:
This is just a small behavior report.

I noticed my capture shown right alignment,
I have 52a056e and ensure the center alignment code on UI.

this was happen when I type something long text (e.g "pppppppppppppppp") and restart JamTaba
and edit by Ctrl-A then the text input cursor was on right. not happen again re-edit time.
look like I took the capture at editing time, center alignment is worked ok.

editing-right-align

teamikl:
@elieserdejesus 
I had similar work about fixed size layout long time ago,
for my case it was for print to paper, that must view all letters, so I had to take ad-hoc approach,
like measure the font width and set smaller font if it was over field.
(it's lower level rendering, e.g. override paint() for Qt. can get fontMetrics info)

for GUI, long text is shown with snip is a general case.
and it's not so high priority to do such lower layer works.
for the layout flexibility too, fixed size layout may be problem on another devices.

This is an Architectural issue, the track is Vertical(so width is narrow) but show Horizontal text.
other places Remote Channels, Chat Message also may have similar issues.


I think this PR can be merge as improved the topic stuff.
(allow still possible be behind, but at least it became more rare case)

elieserdejesus:
@teamikl , I pushed a new commit imrproving the question mark operator (using StringLiteral).

> View Mode -> Mini View does effect to the name input field width?

No. The idea is not show the 16 max users characters everytime in every layout, this is not possible 
without use more width in the local tracks. The local track is fixed width, so big names using large characters (like M) will overflow and some characters will no appear. This is the expected behavior at moment.

> how about set something monospaced font ?

Monospaced fonts will increase the small names width (space usage), right? Maybe using monospaced font common names like "Doublebass", "Andrey_gen" or "HammerOn" have a higher chance to be (unecessary) cropped.

> allow still possible be behind, but at least it became more rare case

Tea, I'm using a little trick to avoid the "text behind" problem. I'm using a QWidget as a conteiner for a QLineEdit and a QLabel. The QLabel instance is used only to show the pencil icon, and the QLineEdit is transparent and without borders. The stylesheet (gray background, rounded borders) is applyed in the "conteiner", not in QLineEdit. So, it's look like a QLineEdit with a small icon in right side, but is not, the icon is a separated widget. Because the horizontal layour in the container the QLineEdit and the icon QLabel will never intersect, and is not possible see the user name text **behind** the pencil icon. Off course the big user names text will be cropped, but this is the common behavior in every text input fields.

teamikl:
@elieserdejesus 
seem I misunderstand this PR topic, so this PR changes was:
[ user name [icon] ]  -> [ user name ][ icon ] ?

> Monospaced fonts will increase the small names width (space usage), right?

Yes. cons is require more width.  this is now out topic though.
Monospaced fonts were useful for fixed size layout, can reduce test cases cost.
non-monospaced fonts does not know max width,
that have to find the most wide alphabet for the font.
its all fixed size for monospaced font.


> this is the common behavior in every text input fields.

Yes, agree. it's an acceptable behavior.

If topic was show all user name characters,
I was going to suggest solution have a horizontal widget (like header/menu etc).

elieserdejesus:
@teamikl 

>  [ user name [icon] ] -> [ user name ][ icon ] ?

Exactly, you got the point. Sorry if my english was not clear enough in the last comments.

> If topic was show all user name characters ...

Tea, I think is not the case at moment. The user name "crop" problem disappear when we have a subchannel (local tracks will user more horizontal space) or a new channel. So I think is not necessary show all user name characters all the time. I think the priority for horizontal space is the "remote tracks panel", to avoid horizontal scrolling during a jam. If we add more horizontal space in local tracks panel (to show all user name characters) we will reduce the horizontal space for remote tracks and possibly (in mini mode) imposing more horizontal scroll usage. As you know scrolling is a bad experience during a jam.

PR #323, opened 2016-04-18
Fixing copy/paste in user name

elieserdejesus:
Avoiding blank spaces in user name when copy/pasting.


jonjamcam:
Tested. Spaces replaced by underscore now when pasting.

PR #341, opened 2016-04-22
fixing-Transmitting-text

jonjamcam:
I'm sorry I didn't figure out this one before :(

image

image

This means correct this slot in all locale?

### DIFF
--- a/src/Common/gui/LocalTrackGroupView.cpp
+++ b/src/Common/gui/LocalTrackGroupView.cpp
@@ -47,7 +47,7 @@ void LocalTrackGroupView::updateXmitButtonText()
         xmitButton->setText(tr("X"));
     }
     else{
-        xmitButton->setText(isPreparingToTransmit() ? tr("Preparing") : tr("Transmiting"));
+        xmitButton->setText(isPreparingToTransmit() ? tr("Preparing") : tr("Transmitting"));
     }
 }
 


elieserdejesus:
Merged in master @jonjamcam , thanks!

teamikl:
I note this for share my issue.
I wanted this PR contains update the *.ts files too

for my translate-ja_JP branch case
- git merge master met conflict
- lupdate with -obsolete  to get the fixed *.ts 

or

I think @elieserdejesus could fix *.ts on upstream easy and safely.
this typo case does not change translation meaning, so it's not translation task.
it's the original resource text only.

this will work on git console (or GUI solution, you can use editor's replace)
```sh
  > cd ./translations
  > `sed -i 's/>Transmiting</>Transmitting</g' *.ts
  > git diff
```

NOTE: the open/close <> in pattern is for exact match.
NOTE: this sed command replace *.ts files directly without backup,
we have git always it can back by git reset --hard if git diff shown unexpected results.

elieserdejesus:
@teamikl , your idea is change (I will change) the TS files directly and push in master, right? No need to wait translators?

teamikl:
@elieserdejesus 
if there was not translated file, then you may need to wait translator.
if someone work with the typo source, then fix typo on master now will have conflict.

the idea is fix typo on upstream, no need to repeat update ts files again.
you could fix the typo after all translation task done.

elieserdejesus:
@jonjamcam and @teamikl , the "transmiting typo" are fixed in master.

jonjamcam:
nice trick!:+1:

PR #348, opened 2016-04-25
Fix Japanese translation for "T"

teamikl:


translations/ja_JP.ts
```diff
@@ -44,7 +44,7 @@
     <message>
         <source>T</source>
         <translatorcomment>表示スペースの都合上、省略文字を利用</translatorcomment>
-        <translation>翻訳</translation>
```
elieserdejesus: @teamikl , you test this translation text? I'm not tested yet (not in home at moment) but I think the small translate button is fixed width, and maybe the string "翻訳" will be cropped.

teamikl: I forgot to explain detail, this patch fix the string cropped
the text on red line will be replaced to green line text.

elieserdejesus:
ops, my bad Tea! I was looking the wrong side in github diff (using splited windows here).

Issue #353, opened 2016-04-28
Show english language names too in language menu


elieserdejesus:
@teamikl proposed:
> hoe about use the menu item field like "日本語 (Japanese)"

@teamikl , can you assume this task? If yes please check **MainWindow::initializeLanguageMenu()**.

teamikl:
@elieserdejesus 
I think there is better quick way,
you can do this task with commit --amend on the branch you already made.
- change translators name and variable name
- commit --amend with new commit message
- rename branch name then PR

you already wrote the code 343787a1c06579a7f22f8adbdf11c4d9d2f3a470, no need to repeat similar task.
If I copy and paste the code, it will show I am the author for the code you wrote.

Instead, I will assume about box.

elieserdejesus:
@teamikl , I pushed the 343787a1c06579a7f22f8adbdf11c4d9d2f3a470 directly in master. If I run a --ammend in this case I will rewrite the master history.

I will do this task soon.

teamikl:
Oh sorry then that scenario does not work.
the commit was already in master I thought PR topic branch.

I can't find the related topic issue for 343787a1c06579a7f22f8adbdf11c4d9d2f3a470
it was a commit only?

elieserdejesus:
> I can't find the related topic issue for 343787a   it was a commit only?

Yes Tea, just one commit. This is why I put directly in master, but was a bad idea :)

elieserdejesus:
@teamikl , just to confirm, you mentioned
> Instead, I will assume about box.

The idea is an "About ..." entry in the menu? Or the idea is put a entry in menu to link to "translators section" in Jamtaba website?

teamikl:
@elieserdejesus 
I did not think about link at the moment
let's move this topic to 343787a

elieserdejesus:
Ok @teamikl , I will assume the "translators link" task.

elieserdejesus:
Done in #355.

PR #355, opened 2016-04-28
Showing english language names

elieserdejesus:
Solving the issue #353

@teamikl , I'm capitalizing the language names. I checked the japanese symbol and it's the same for capitalized and non capitalized. Can you confirm?

image


teamikl:
>  I checked the japanese symbol and it's the same for capitalized and non capitalized. Can you confirm?

Yes, This is ok. the most of Japanese letters does not have cases.

----
I have code review about the capitalize function namespace.

that function should belong to utility namespace.
If you wanted temporary function (or macro) there, c++11 has support lambda.

```cpp
  auto capitalize = [](const QString &string) {
    return string.left(1).toUpper() + string.mid(1);
  };
  qDebug() << capitalize("test");
```

demerit is, can't make unit test of the function.
abuse lambda make debug hard, but ok for this simple case.

teamikl:
I forgot to mention compiler support  (I did not test msvc)
I think that lambda snippet work unless compiler too old.

elieserdejesus:
@teamikl 
> that function should belong to utility namespace.

I choose the namespace approach, I think the capitalize function can be reused in future Gui features.

PR #360, opened 2016-04-29
[WIP] Fixing linux installers

elieserdejesus:
- [x] generate different files for 32 and 64 bits
- [x] include the Qt libs in distribution
- [x] Create a script to set the library path (by default linux will not use qt libs in distribution dir)


elieserdejesus:
I Finally get a linux installer working fine in Ubuntu 32 bits:

screenshot from 2016-04-30 17 43 01

Issue #369, opened 2016-05-08
Gain buttons in horizontal tracks layout in wrong position


jonjamcam:
I find that the boost/att direction should follow the level meter as it does in the vertical tracks layout, but here it's reversed:

![image](https://cloud.githubusercontent.com/assets/15310433/15098499/a1b62476-1515-11e6-818b-af7e1cecd61c.png)

I always get it wrong when changing this setting during jams

jonjamcam:
@elieserdejesus I thought this one was easy to solve, but when I changed the buttons order, the vertical layout is changed too, so I closed the PR https://github.com/elieserdejesus/JamTaba/pull/370.

elieserdejesus:
@jonjamcam , check the file **NinjamTrackView.cpp**, like **151**:
```
boostWidgetsLayout->setDirection(QHBoxLayout::LeftToRight);
```

Try change the layout direction to **QHBoxLayout::RightToLeft**, this change will be applyed only in horizontal layout.

jonjamcam:
Solved in PR https://github.com/elieserdejesus/JamTaba/pull/371. Closing...

PR #370, opened 2016-05-08
Changing the gain buttons order

jonjamcam:
This should correct https://github.com/elieserdejesus/JamTaba/issues/369


jonjamcam:
this pr changes the vertical layout too. Bad solution. Closing

PR #376, opened 2016-05-12
Emoticon for chat message

teamikl:
screenshot 8

This is the first step, support emoticon on chat message.
the notation is `:smile:` some of the icon identifier are long.
this commit does not implement the emoticon name completer, yet.

I post this PR for starting review.
notation and icon, size etc

I resized 64x64 to 24x24 then total size of icon png files is about 1.1MB
## Emoji emoticons

How many emoticons supported,
documented here http://www.emoji-cheat-sheet.com/
## Known issue

Translate change like this:

screenshot 10
screenshot 9
## More tests

I did not test chord progression parser.


jonjamcam:
just tested, looks fantastic!, @teamikl very nice start for the implementation!
:smile:

teamikl:
I should made separate commit to add too many png files, that made diff and commit log complex.

Implementation is very simple, only the EmojiUtil::emojify
added two tests tests/manual/gui/chat and tests/auto/gui/emoji

about emoji resources, there is the most important topic, it's license.
emoji data is part of unicode, but this glyph are seem Apple's.
as I saw in http://www.emoji-cheat-sheet.com/ list, used in open source projects.
but I am not sure what license.

elieserdejesus:
@teamikl and @jonjamcam , what you thing about reduce the emoji set? I thinking people use just a little subset of emojis, maybe we don't need tons of emojis, just a small and usefull/meaninfull subset. 

Check this: http://emojitracker.com/

elieserdejesus:
Another idea: try "imitate" twiter and facebook. How the emojis are typed/inserted in these tools? People are very used to these tools, it's very interesting use the same emoji codes, no need to learn a new emoji "syntax".

teamikl:
@elieserdejesus 

I looked facebook's the chat interface support emoticons
listed on a grid but the different is numbers of items.

reduce emoji set is good for the input, it will be easy to look up.
another effect is reduce the resource size, this should be discussed
embed emoji set in resource or load icon from external site.

negative side is maintenance non-standard is a cost.
Example, needdocumentation, since it will be own subset. and Problem scenario
using subset and embed's demerit is if added new icon and JamTaba version mismatch,
they may show broken image.

> no need to learn a new emoji "syntax".

good point.

about the syntax, it had problem in translation.
I was thinking use unicode emoji data characters, they are translation safe.
non-JamTaba utf-8 client also can see the emoji glyph if they had a font.

screenshot 2


another service using same emotions, Slask had a modern interface, it too rich ?
screenshot 1

teamikl:
This is how input emoji in slack, they support two ways.
no need to learn the syntax if use the insertion tool. the syntax is an optional.

screenshot 3

screenshot 4

`:emoticon-name:` was standard syntax for emoji,
there is implementation merit, that simple regex can handle various names.
no need mapping dictionary for custom emoji set. as the result [emojify function](https://github.com/teamikl/JamTaba/blob/f331a6ed1f5af1245be7797587bbe57021c5b5b5/src/Common/gui/EmojiUtil.h) became so simple.

custom way may bring another issue,
e.g. when add new emoticons, need discussion every time have to care the chords parse.
and also if conflict use situation, `:|` look like face but someone may want use as repeat bars like
`|: Bbm7 Cm7 | DbMaj7 :| x3   | Abmaj7 | F7(b9) |` (assume non-JamTaba user post chords like this)

elieserdejesus:
@teamikl  

> this should be discussed: embed emoji set in resource or load icon from external site.

I think use a embed small subset of emojis is simple, but I understand the problems about version mismatch you mentioned. 
I have another 2 little discussions:

1. I'm a bit worried about auto completer if using emojis from external resources. Using embed resources the autocompletion is perfect (real time), using external emojis we will experience some delay (slow connection in user side, or an overloaded emoji server).
2. Using a embed emoji subset will impose a task for us: how decide the size of the subset? How decide the "best/most used" emojis?

When I think about **2** my first idea is: just use the external resources. They are tested and aproved by the people.  

> about the syntax, it had problem in translation. I was thinking use unicode emoji data characters, they are translation safe. non-JamTaba utf-8 client also can see the emoji glyph if they had a font.

Tea, after look at some [unicode glyphs](http://unicode-table.com/en/#miscellaneous-symbols) they "sound" too simple in comparison with modern emojis. 
About the **translation problem**: the problem is sending the emoji codes to translation service and receiving bad translations? or the problem is receiving translations from service, but loosing the emoji symbols?

> no need to learn the syntax if use the insertion tool. the syntax is an optional.

Shure, I think this is the best way. Can we have a insertion tool too? :smile: 


> :emoticon-name: was standard syntax for emoji,

so, +1 here to use the standard :smile: And I tested the ":smile" right now :+1: .The github completer works nice!

teamikl:
@elieserdejesus 

> Can we have a insertion tool too?

Yes, but not on this PR 
I will separate PR completer and insertion tool to discuss different UI
for now, I am stuck on QLineEdit/QCompleter internal event.

## Embed or External
I will make simple test if QTextEdit allow img tag load external image asynchronous  
if not then need customize or use webkit component I feel it's heavy.

## translation issue
it was issue for the "syntax" `:guitar:` -> `:ギター:` (not show icon when translated)
translator did not touch emoji code `🎸` this unicode character was translation safe.

I did not mean use the glyph instead of icon, use the character as marker then
non-JamTaba client which not support emotion but still can see the glyph.
this is just idea stage, that has side effect, this way assume utf-8 so
I think NINJAM/reaNinjam on window they will see broken font glyph. (not tested)

## using unicode emoji character as the marker for emoticons
it will be translation safe and another effect
using the unicode emoji, implementation can be more simple if there was font
because it's only setFont to set emoji font. (not tested too, I don't have the font)

teamikl:
@elieserdejesus 

> I'm a bit worried about auto completer if using emojis from external resources. Using embed resources the autocompletion is perfect (real time), using external emojis we will experience some delay

if loading many image files use many connection and maybe delay,
Modern web does some optimization, use font data
(load single file and showing chunk for icon) like [font-awesome](http://fontawesome.io/) does.
we can pre-load then no delay but it will be almost same as embed.

teamikl:
> we can pre-load then no delay but it will be almost same as embed.

there was little different, pre-load at runtime can solve the version mismatch issue.
it will be possible to update.

elieserdejesus:
@teamikl 

> need customize or use webkit component I feel it's heavy.

Please avoid webkit. It's very huge, I will need recompile the static Qt libs here just to use a emoji.


> I did not mean use the glyph instead of icon, use the character as marker then non-JamTaba client which not support emotion but still can see the glyph.

Ah, my bad. I see now. But I think you are right about other ninjam clients (the unicode glyph will not be recognized). I think is better receive a *"Hi all :smile"* than a *"Hi all :some-strange-code"*.  


> pre-load external resource at runtime can solve the version mismatch issue.

I agree, pre loading will be nice.

teamikl:
@elieserdejesus 

> Please avoid webkit. It's very huge, I will need recompile the static Qt libs here just to use a emoji

OK, but unfortunately, ` Using a embed emoji subset will impose a task for us: how decide the size of the subset? How decide the "best/most used" emojis?

if need to select subset then
Following standard or another popular service doing is often easy solution.

If using external resource then full set can be an optional,
but if pre-load subset then task again for the select of subset.
another combination is pre-load full set (e.g. web-font) and provide subset.

elieserdejesus:
@teamikl 
>  the installer file size (emoticons resource size is about 1MB I reduced)

1MB is ok, these imagens will be compressed by installers. The final size is will be ok.

> another combination is pre-load full set (e.g. web-font) and provide subset.

I like this option, and use a subset used in popular services.

teamikl:
@elieserdejesus 
> I'm using QLabel in chat messages and QLineEdit in chat input. Where are you using QTextEdit? Emoji completion?

QTextEdit was my local test code. However the limitation was same,
QLabel also use QTextDocument internally which QTextEdit use.

`img` tag for external url does not work,
possible solution is using unicode emoji character and
 download external font resource and setFont 

> 1MB is ok, these images will be compressed by installers. The final size is will be ok.

can't expect much compression here, tho. image were png which already compressed.
another size effect is git repo size, each files are small but many files,
they should be into repo? no need to manage version if won't update.

----

I schedule my tasks

1. embed full emoticons and `:smile:` syntax (this PR)
2. insertion tool (discuss subset here)
3. completer for the syntax

- unicode character and pre-load web font
  I had not try this on Qt, need to make proof of concept first.

teamikl:
> img tag for external url does not work,

another way was embed base64 data `<img src="data:image/png;base64,..."`
It's worked in QLabel::setText.


I should correct information, I though it was rare to update emoji data,
but they add new emoji in few year. http://unicode.org/emoji/charts/emoji-versions.html

teamikl:
I report that tested embed emoji font (.ttf file)
embed the full set font data in resource was not good result.

resource compiler generated qrc_xxx.cpp file (60MB cpp file for the 10MB ttf data)
that represent resource data as unsigned char table,
then compile resource would use much time to parse the source.

so embed also had two ways, embed into exe or embed in installer
install into to user's local directory. (img tag works for local path)

- resource: embed or external
- show icon:  img tag (resource/local file/embed base64) or setFont/unicode
- emoticon syntax: safe for chord parser/translator/other clients

set emoji font approach did not success, yet.
and as I read document that has platform issue on linux/fontconfig.

elieserdejesus:
> ... can't expect much compression here, tho. image were png which already compressed. another size effect is git repo size, each files are small but many files, they should be into repo? no need to manage version if won't update.

I'm seeing you are testing possibilities at moment, is better define the implementation approach first and decide where the resources will be stored later.

> resource compiler generated qrc_xxx.cpp file (60MB cpp file for the 10MB ttf data)

Now I'm scared :)  This means 10 MB embed ttf in the project?

elieserdejesus:
hi @teamikl , any news in this PR? I suppose you are busy at moment, I will consider schedule this PR to the next release, no worries.

jonjamcam:
@elieserdejesus I think Tea is not finishing this PR (It's been a year now). Do you think is possible to implement a simple "github style" drop menu and have it implemented for 2.0.20?

image

elieserdejesus:
@jonjamcam , I will look at this after finish the looper tasks, will be nice have this feature in chat.

I think a pending question is "what non JamTaba users will see when we send emoticons in chat?" This can be more trick to solve, Tea was working in some ideas, but I can't remember the details.

jonjamcam:
> what non JamTaba users will see when we send emoticons in chat?

I tested it and others see **:emoticon-name:**. I don't think there's much to do about other clients, just like the chords feature. Do you think @elieserdejesus this is not good behavior?

elieserdejesus:
> Do you think @elieserdejesus this is not good behavior?

I think this can be good to make people curious about the feature and use JamTaba 😄 . But for other clients the chat will be more poluted with these :emoticon-name texts.

Another idea is reduce the emoticon set to just a few "text translatable". For example, we can replace **:emoticon-smile** with just **":)"**. JamTaba users will see a emoticon, other clients will see **":)"** in chat. The **+1** and **-1** are another easy to "translate" examples. 
This idea will work with a small emoticon set only, but I think can be good enough.

jonjamcam:
> Another idea is reduce the emoticon set to just a few "text translatable". For example, we can replace :emoticon-smile with just ":)". JamTaba users will see a emoticon, other clients will see ":)" in chat. The +1 and -1 are another easy to "translate" examples.
This idea will work with a small emoticon set only, but I think can be good enough.

I agree. That was actually the idea in the beginning. I found a chart with some common ones:

jonjamcam:
@elieserdejesus Utf-8 support emoji, so use this fonts to transmit them over the chat looks to me the correct approach. Tea and you were already discussing this issue before?

I found this http://www.guigarage.com/2015/01/support-emojis/ with details.

> Emojis and unicode
Let’s start with a fact that is a big benefit when working with emojis: Since 2010 all the emoji icons that you know from your favorite chat application are standardized in the unicode standard. The unicode standard defines more than 700 emojis. This means that each emoji is defined as a unicode character and has its unique unicode code:

If I understood correctly, any UTF-8 client can decode the emoji correctly. If not (nijam client, old reaninjam, etc.) a � charcter will appear. If this is true then the compatibility problem is solved. Is this an option in your opinion?

jonjamcam:
> I think a pending question is "what non JamTaba users will see when we send emoticons in chat?"

Now I can see a clear solution for this.

1.-First we need a way to choose a specific emoji from a list (for example Tea's 867 emoji list from this PR). A simple drop down menu, for example, just like github I think is enough.

2.- We can use combinations of unicode special characters like ٩◔̯◔۶ to  "transmit" the emoji over ninjam chat. The receiver detects this maps (that we design) and posts the corresponding emoji.

3.- everyone is ˙·٠•●♥ happy ♥●•٠·˙. No :text: syntax anymore. 

Does is look good to you @elieserdejesus ?

elieserdejesus:
@jonjamcam , this utf-8 approach make sense for me too. Let's work in this issue after finish the looper stuff (I'm working in save/load layers content at moment).

jonjamcam:
> Let's work in this issue after finish the looper stuff

Of course. sry. I was getting ahead of myself. Just documenting for later...

elieserdejesus:
> Just documenting for later...

Thanks @jonjamcam , I appreciate, these details will help a lot.

jonjamcam:
@elieserdejesus  I did some work "translating" the emoji to utf-8 symbols (as commented here: https://github.com/elieserdejesus/JamTaba/pull/376#issuecomment-285301523). I think all these are compatible with Jamtaba:

[emoji translations.txt](https://github.com/elieserdejesus/JamTaba/files/871098/emoji.translations.txt)

Of course there are some missing, but we could use these for a start and add gradually more in each version. What do you think? Do you think there are some important missing?

elieserdejesus:
@jonjamcam , this is very a good emoji set to start. I will work in this feature in next version.

elieserdejesus:
I'm closing this PR. The general idea and some resources from this PR are used in https://github.com/elieserdejesus/JamTaba/pull/973

Issue #377, opened 2016-05-12
Inverse Stereo Image Button for Standalone


jonjamcam:
To invert the stereo image on one specific channel is easy in Reaper (before send to Jamtaba), but there's no way to do it in Jamtaba Standalone and sometimes it's very important when the plugins do not have the option to do it (i.e. left handed drummers/percussionist)

elieserdejesus:
@jonjamcam , the idea is invert stereo only in local channels?

jonjamcam:
>@jonjamcam , the idea is invert stereo only in local channels?

yes. that's the idea. with a button if possible

elieserdejesus:
@jonjamcam , you have ideas for the button text or icon? Something like "LR", and if you click the text change to "RL"?

jonjamcam:
I'm looking into it. I'm thinking about something on these lines:

http://thumbs.dreamstime.com/z/stereo-symbol-10603870.jpg

but maybe with opposing arrows....I'm into it

jonjamcam:
Another idea:

![image](https://cloud.githubusercontent.com/assets/15310433/15259727/5f2b4d76-1929-11e6-8503-c163cb32143c.png)

This is assuming L is blue and R is red.

elieserdejesus:
another one: 
![image](https://cloud.githubusercontent.com/assets/1012741/15259836/e127e4ba-1929-11e6-9dec-fd6ca7f5be4c.png)

jonjamcam:
That above looks really good!

jonjamcam:
I found the Channel Pan control is inverted when the invert stereo button is pressed (need a separate issue for this?)

![image](https://cloud.githubusercontent.com/assets/15310433/15270594/95c9ba3e-19fb-11e6-9a64-50f58565ca06.png)

elieserdejesus:
@jonjamcam , I keep the pan controls in "normal" because I think left side is always left side, even for left handed people. But I'm no shure what is more confuse: keep the pan controls in "normal" mode or change to R -> L. All this inversions are burning my brain.

You idea is flip pan controls entirelly? R -> L?

jonjamcam:
@elieserdejesus . The inversion is supposed to work pre-fader, so it "corrects" a wrong incoming stereo image. All further control in the channel chain should work the same. For example:

a Right handed drummer sets his drum set like this:

[...]

PR #379, opened 2016-05-14
Stereo inversion in local tracks

elieserdejesus:
- [x] Add/create icons for stereo inversion button
- [x] Put the stereo inversion button in local tracks layout
- [x] Disable stereo inversion button if local input is audio and mono.
- [x] Implementing the stereo inversion in audio
- [x] Store/remember the stereo inversion for each local track
- [x] Invert peak meters.

This PR is implementing #377


elieserdejesus:
@teamikl , @jonjamcam proposed a feature to invert stereo channels in local tracks #377.

I'm working on this feature and a new maintence problem appears. In the last commit I added the button to invert stereo in local track layout:
image

But I have to copy/paste the stylesheet code in Flat.css and Rounded.css to have the button looking ok in both themes. Of course copy/paste stylesheets will be a pain for 2 or more themes. @teamikl , have you some good idea about **how maintain these stylesheets without copy/paste?**

The first and simple idea in my mind is just hard code the common stylesheet in C++ side and the specific theme details in .css files. For example, the icon for stereo inversion button is the same in both themes, so this can be setted in the C++ code. On theme is using rounded borders, this can be writed in .css file. I think is not a 100% elegant solution, but is better than copy/paste :)

elieserdejesus:
@jonjamcam , I finished the stereo inversion feature. Can you help testing the feature?

jonjamcam:
@elieserdejesus Just tested the branch and I find a problem with the "Disabling the stereo inversion if input is mono or no-input" commit (https://github.com/elieserdejesus/JamTaba/pull/379/commits/17ef3abb80914ea59bc62080d44400a9b3ea9add)

I think this should be the case only when there's no VST activated because many vst FX convert a mono input to a stereo output.

elieserdejesus:
@jonjamcam , I removed the "no-input" checking in d0aad91. Now the button is disabled only when the input is mono. The input cat be "no-input" but we can use a stereo VSTi generating a loop, for example, and in this case stereo inversion make sense.

elieserdejesus:
@jonjamcam , I changed the stereo inversion to pre-fader. Pan controls are working correctly right. Can you check this branch again, please?

jonjamcam:
@elieserdejesus It's working OK now.

PR #381, opened 2016-05-15
Organizing stylesheets

elieserdejesus:
- [x] Split stylesheets (CSS) in small files.
- [x] Separate common stylesheet in different css files to avoid duplicated code between themes.
- [x] Load the themes using the new file structure.
- [x] Build the list of themes using the new resources file structure.
- [x] Reorganize and document the css files


elieserdejesus:
@teamikl and @jonjamcam , I'm working in a new structure for themes. I splited the big CSS theme file in small files and created a directory structure to organize "common" and "theme specific" code. All shared CSS code segments are in "common" folder. In "css/themes/Flat" are the CSS segments for "Flat" theme. The same for the Rounded theme.

The manitence of stylesheets is very improved now. No more copy paste and is easy navigate in the small files.

The last task is clean some useless code and write some comments to help others understand the connection between CSS and Jamtaba GUI elements.

jonjamcam:
I see it's much easier to go through the parameters now if you want to create a new theme. So this is the full tree now:

image

  So I see each new theme should have nine new files under /themes dir. Creating these 9 files should be enough to have a new one?

jonjamcam:
I created a prefix for a black theme and copied the files from flat, but although the theme DOES appears in the menu, I have this error in the compiler:

image

and the result is:

image

elieserdejesus:
@jonjamcam 
> So I see each new theme should have nine new files under /themes dir. Creating these 9 files should be enough to have a new one?

It's not necessary create all 9 files. The 'common' is always loaded when a theme is selected. You can create a theme and 'override' only the Chat.css, for example to have different Chat colors. 

@jonjamcam about your error: Click with right mouse button in jamtaba.qrc file (the resources file) and choose the option open in "plain text editor". The file will be opened in QtCreator as a text file. Check the entries for the Flat theme. All files are using an **alias** attribute. You need these aliases your yout Black theme too.

I added a new commit cleaning and documenting the 'common'. Maybe the documentation can help you. I will do the same job in flat and rounded stylesheet soon.

jonjamcam:
> Click with right mouse button in jamtaba.qrc file (the resources file) and choose the option open in "plain text editor". The file will be opened in QtCreator as a text file. Check the entries for the Flat theme. All files are using an alias attribute. You need these aliases in your Black theme too.

that solved it! THX. I will work on a black theme.

elieserdejesus:
Done!

Issue #306, opened 2016-04-13
Implement a "Mute Chat" button


jonjamcam:
![](https://plnami.blob.core.windows.net/media/2015/03/troll031615-800x500.jpg)

I'm thinking about a simple and fast troll block button. The idea is that when a troll is detected in a room (chat only or chat/track) we can make him vanish locally (from our client only) so his audio/chat doesn't bother us anymore .The problem with trolls is that they tend to tarnish the mood and break concentration and focus, so my approach is to have 2 levels of banning.

1.- light: The chat is put on dark color and the channel is grayed out and muted for the ongoing session only.
2.- full  : the chat/track is not shown for this individual anymore (including future sessios) until the user unbans him willingly (i.e. opening a menu, or something).

Before I go on I want to ask you @elieserdejesus and @teamikl if you think this is an important topic to adress in the Jamtaba implementation or not?

elieserdejesus:
@jonjamcam , personally I never had problems with trolls in ninjam. But we known very well the ninjam troll legends :) And we have another cases every time related in ninbot forum. Sometime ago I have a similar idea to block the "troll" messages in chat. In general I think this is a "sad feature", but I have to admit this will be usefull for many ninjamers. 

I think we can discuss this feature. Spend time discussing "trolling" It's not my preffered topic, but it's necessary :(

About implementation details: we need block the user using the USER IP. But a "true troll" knows about proxies, and we are blinded again if the IP is different. It's impossible block a smart troll :)

teamikl:
@jonjamcam 
it's difficult for NINJAM public servers. because the most users are anonymous.

>  a troll is detected in a room (chat only or chat/track) 

e.g. if long loud noise was detected as troll action,
but if it was unexpected behavior by new comers with wrong setting,
they does not know how to setup yet.
someone should support or they may give up to use the software.

> his audio/chat doesn't bother us **anymore** .

How to Identify the troll user?
they can type any name and IP address may be changed.

use IP address for identify has rare issue,
it's because the last field is masked,
different people who live in same building had same address string.
If the troll spoof other user and get banned intentionally... should avoid that.


I can understand the case you may assume, someone who muted(or uncheck recv)
but they try reconnect (mute will be unchecked by reconnect).
There is no perfect way to block them, but still there are several useful cases.

jonjamcam:
@teamikl 

> different people who live in same building had same address string.
If the troll spoof other user and get banned intentionally... should avoid that.

true.

[...]

PR #385, opened 2016-05-21
[WIP] Chat mute feature

elieserdejesus:
- [x] Implement the chat message block per user
- [x] Decide if users will be blocked just for the current servers or for all servers.
- [x] How Jamtaba users will block another user in chat?
- [x] Implement the context menu to block and unblock users in chat

Implementing #306


elieserdejesus:
@jonjamcam and @teamikl 

I started the "chat mute" feature and I'm a bit stucked. **How Jamtaba users will block/mute another users**?

I have some ideas, but all ideas are not really good at moment:

1 - Put a button with a fancy icon in bottom of another buttons in the tracks (mute, solo, boos, etc). This was my first idea, but now I realized the chat mute is a **per user** feature, not **per track/channel**. I'm trying to decide a good place for this chat block/mute button, but no good ideas at moment.

2 - Put a little button in the chat message, like the translate button. When you receive a chat message from a troll you just click this button and the troll is blocked. I think this solution is good enough, but how unblock users?

3 - Unblock is really necessary? If not (I think is not) the 2nd idea is enough for me.

**Another topic**
As we discussed before in #306, is not possible save/remember the "trols" because IPs can be changed using proxies, etc. So, my last topic is: the user will be blocked until Jamtaba is closed or just in the current server jam?  My first commit is blocking just in the current server, but I think make more sense block the troll in general (all servers).

jonjamcam:
I like idea 2 too. it's enough and easy.

>but how unblock users?

Because the block is associated with a username, then the username in the mixer could change color (red for example) or _italics_. So you could right click on the italics user and choose unblock, for example

elieserdejesus:
ok @jonjamcam , I think a "context menu" (activated with right mouse button) will solve the "unblock" question.

jonjamcam:
> the user will be blocked until Jamtaba is closed

I think this is the best option. In my experience, trolls do not like the job too long :D. Even for professional trolls (who I've seen) they change the username from one day to the next.

elieserdejesus:
@jonjamcam , I finished the chat mute feature. Tested using 2 local users here.

There is a little problem with the current implementation: users are blocked by name. The names are unique inside one server (ninjam servers don't allow duplicated names), and the current solution will work ok. But if you block an user called "elieser" in ninbot.com:2049, jump to another server and there is a different user in this server but using the same name this user will be blocked, and this result is wrong.  This is a limitation of the current implementation, but this cases (same user names in different servers) are rare cases.  If you have enough problems in future I can try another approach using the user IPs.

jonjamcam:
> this cases (same user names in different servers) are rare cases

I agree. The case when you block a troll in one server and the same name is used by a normal user in another server is not likely at all. Anyway this can by corrected unblocking the user in the new server or restarting Jamtaba (both easy solutions).

> ninjam servers don't allow duplicated names

I was not aware of this. Maybe it's a recent implementation and/or it has exceptions. I have jammed recently (not more than 1 month ago) with two "testing" and "anon" in the same room.

jonjamcam:
I tested the block button. The only problem I have is the location. It's too near the translate button. For recurrent translate use it could block instead by accident.

@elieserdejesus  What do you think about this location?

image

elieserdejesus:
@jonjamcam 
> ... For recurrent translate use it could block instead by accident.

I agree, block by accident is a real issue if the buttons are small and closest.

Your 2nd idea is interesting because the button is close to username, meaning the "user" will be block. Let's adopt this position (in left of user username).

jonjamcam:
@elieserdejesus the button is not appearing anymore after your last commit in my system. Is it just me?

elieserdejesus:
@jonjamcam , tested again and is working. The block button is hidden in server bot messages and in your own messages.

jonjamcam:
ahh ok :D

jonjamcam:
one question a message is shown "XXXX blocked from chat" Is this message visible to all users?

elieserdejesus:
>  Is this message visible to all users?

No Jon, this is a local only message. When I readed your question I was not 100% shure, but tested again and it's just local. :)

PR #393, opened 2016-05-25
[WIP] Rms metering

elieserdejesus:
- [x] 	Computing RMS values for audio bufferr
- [x]  Showing both peak and rms in meters
- [x] Drawing a rms mark/line to identify the rms peak
- [x] Disable/enable rms and peak metering in tracks context menu.
- [x] Improve the rms window size to use samplerate as base parameter.

### DIFF
--- a/src/Common/MainController.cpp
+++ b/src/Common/MainController.cpp
@@ -52,6 +52,11 @@ void MainController::unblockUserInChat(const QString &userNameToUnblock){
 
 void MainController::setSampleRate(int newSampleRate)
 {
+    foreach (Audio::AudioNode *node, tracksNodes.values()) {
+        int rmsWindowSize = Audio::SamplesBuffer::computeRmsWindowSize(newSampleRate);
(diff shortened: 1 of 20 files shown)


elieserdejesus:
The result or rms and peak metering:
image

elieserdejesus:
@jonjamcam , I finished a new fancy feature, a RMS metering. Is integrated in master now.

jonjamcam:
ok. I just tested it minutes ago from the branch. nice!

NakajimaYusuke:
Thank you for the the great work!

I think RMS meter is enough decent alternative for loudness meter.
Definitely, it should be a good volume indicator than the peak meter.

Once you have determined the optimal RMS value,
I hope that the recommended RMS value is clearly displayed on the meter.
I think "how to indicate the recommended value?" is also important point.

This is a little example.
T-RackS CS Metering (IK Multimedia Production)
ikmeter
Red and green line shows the recommended value range.

Issue #308, opened 2016-04-13
BPI/BPM feedback needed inside the server


jonjamcam:
This is an important detail. As we all know the only feedback we have of what BPM/BPI is the server currently using is the BPM/BPI combo boxes.

 So what happens in those cases that we type a value in the box to vote, but  the vote does not reach majority is that we are stuck with a suggested BPM/BPI in the combo box, but the actual BPI/BPM didn't change, so now we don't remember what values we are dealing with. The solution seems simple to me:

![image](https://cloud.githubusercontent.com/assets/15310433/14503266/b24b6bbe-0185-11e6-8730-bc1c692467be.png)

elieserdejesus:
Good idea. 

I think at moment the combobox value is not changed/undo when the vote not reach majority. We can think in a solution for this problem too. The difficul in this case is **how we know the voting was aborted**? 

**A first idea**: the "voting" chat messages text contains the expiration time. We can use this time value, and if this time is expired before a change in BPI or BPM the combobox value is setted to the current BPI/BPM. 

**a second idea**: If I remember correctly the expiration time is 60s (configurable in server?). In this period (60 s)  users will see a "wrong" value in the combobox. I think use a different color in combobox text is interesting. For example, while waiting for votes the combobox text color will be yellow (same color of transmiting button in "preparing" state). When the votes are confirmed the combobox text color is changed to the "normal" color.

jonjamcam:
>the expiration time is 60s (configurable in server?)

yes. Here's the section of the ninjamserver.cfg file that determines that:

![image](https://cloud.githubusercontent.com/assets/15310433/14504372/f1acc438-018a-11e6-880a-ca7352436096.png)

The second idea is great! :+1:

PR #395, opened 2016-05-26
[WIP] Bpi and bpm feedback inside ninjam window

elieserdejesus:
- [x] show current bpi and bpi in top of ninjam window
- [x] use red color in bpi and bpm combos when a votation is running.
- [x] detect when a votation expires and reset the combo values and color.
- [x] Disable voting buttons in chat when votation is expired.
- [x] Handle double expirations (user vote a bpi change and after some seconds vote a BPM change).
- [x] Combo is not backing to normal state when vote is accepted. 
- [x] Red color is OK, but the combobox should not be replaced by text-only. At moment is not possible change the vote to replace a wrong vote value until the previous vote expires.
- [x] Expanding the local track (in mini mode) the username and BPI, BPM texts are overlapped.

Implementing #308.


elieserdejesus:
@jonjamcam , I finished all planned taks in this PR. I write some auto tests to check valid  and invalid voting messages and tested in my local ninjam server too (Reaninjam and Jamtaba). Can you test too? The main changes are:

* Now we have BPM and BPI in top of ninjam window.

* Feedback in bpi/bpm combos while a voting is running. I used the red color, yellow not working, dificcult to read.

* Combos color and values are reseted when a vote is expired. I changed the server config file to use a small expire time to run my local tests.

* Now the vote button in chat messages appear only when another user is creating the votation. If you create a votation (!vote bpi 32) the vote button will not appear because you can't vote again using the button.

* Hidding the vote button when the vote is expired. After the expire time, if the user click in the button the server will reject the vote, so no make sense keep the button.

jonjamcam:
Did a quick test and found two issues. I was able to test using two computers, so I changed bpm in computer test1. All good, no button appeared in test1. Computer test2 had the voting button. As expected, the Combobox for BPM was replaced by red text before I accepted the change on computer test2.

image

1.- Now when the vote is accepted and the bpm is changed (picture above), the red text was supposed to become again a combo box, which didn't happen.

2.- The other problem in mini mode happens when you expand the local track and use the maximum width, the username and BPI, BPM texts are overlapped (see picture above). I guess this will happen too with long usernames. One idea I have is to hide the "connected as...." when expanding the local track in mini mode. This way you will still have the username information in the upper left corner.

image

When local Track is collapsed, you get the information as it is implemented now.

elieserdejesus:
@jonjamcam 

> ... I think once the new value is changed the combo box should return to normal inmediately or when the next cycle starts (next beat 1).

Understand. At moment the combo is not backing to normal state when vote is accepted. But the last part is implemented (when the next cycle starts).

> I think the red color is OK, but maybe the combobox should not be replaced by text-only.

Totally agree. At moment is not possible change the vote to replace a wrong vote value until the precious vote expires.

> 2.- The other problem in mini mode happens when you expand the local track and use the maximum width, the username and BPI, BPM texts are overlapped.

Ahhh, the mini mode is always a problem :)

elieserdejesus:
@jonjamcam , about the problem in mini mode: 

I dropped the "Connected as" text in mini mode, but the user name is still cropped when the user name is big:
image

I have no other ideas to show all infos without cropping. I can see 3 routes now:
1 - Keep the user name cropped.

2 - Drop/hide the bpi/bpm info in mini mode. Bpi and bpm are showed in the combos, and now we have some feedback when the combo values are diverging from current server values. Maybe the combos are enough.

3 - Drop/hide the user name in mini mode. The user name is always showed in the left (local tracks).

What you think about?

jonjamcam:
I have this idea, but I don't know if the layout makes it possible:

image

Use a double line display.

elieserdejesus:
> double line display 

Good idea, I will try this.

elieserdejesus:
@jonjamcam , the "double line" idea solve the problem:
image

All review tasks are done. Can you run a last test before integration?

jonjamcam:
Tested. Working fine.

 One thing I noticed is that if you change your mind, for example and vote twice, then someone accepts the first vote, it will ask **you** to vote the old BPM, instead of applying the first submitted vote. I think this is the expected thing to happen because the first vote was a discarded choice.

 So IMO it's working perfect!

PR #397, opened 2016-05-26
muted basetrack color fix

jonjamcam:
@elieserdejesus Track layout don't have problems here:

image

The solution in this case was simply to use the Flat theme code that was working and use the original rounded gradient that was there. Can't tell why this was hapenning, just did a quick fix keeping the intended colors.

### DIFF
--- a/src/resources/css/themes/Rounded/BaseTrack.css
+++ b/src/resources/css/themes/Rounded/BaseTrack.css
@@ -96,11 +96,12 @@ BaseTrackView
    font-size: 8px; /*db value */
 }
 
-BaseTrackView[unlighted="true"]             /* track when is not xmiting */
+BaseTrackView[unlighted="true"],                            /* the track */
+TrackGroupView[unlighted="true"] > #topPanel                /* the track group top panel: used to show user name */
 {
     background: qlineargradient(x1:0, y1:0, x2:0, y2:1,
-        stop:0 rgb(150, 150, 150),
-        stop:1 rgb(90, 90, 90));
+    stop:0 rgb(150, 150, 150),
+    stop:1 rgb(90, 90, 90));
 }
 
 PeakMeter


src/resources/css/themes/Rounded/BaseTrack.css
```diff
@@ -96,11 +96,12 @@ BaseTrackView
    font-size: 8px; /*db value */
 }
 
-BaseTrackView[unlighted="true"]             /* track when is not xmiting */
+BaseTrackView[unlighted="true"],                            /* the track */
+TrackGroupView[unlighted="true"] > #topPanel                /* the track group top panel: used to show user name */
```
elieserdejesus: @jonjamcam , this change will affect the topPanel background, not the tracks background. This is correct? Is your intention?

jonjamcam: > This is correct? Is your intention?

Not at all. Maybe this was a bad solution. I tested some other possibilities, but none worked. I just copied the code from Flat theme and added the gradient for Rounded, but this is obviously changing other things.

The original fix should have been to change only the values in the code for `BaseTrackView[unlighted="true"]`, but I tried and there was no change.

jonjamcam:
I think I understand now what happened. The changes I tried before didn't work because I didn't use clean. 

So now i'm taking the long road to be sure every change I make is correctly applied.

jonjamcam:
@elieserdejesus Changing the code: 

`BaseTrackView[unlighted="true"]             /* track when is not xmiting */
{
    background: qlineargradient(x1:0, y1:0, x2:0, y2:1,
        stop:0 rgb(150, 150, 150),
        stop:1 rgb(90, 90, 90));
}`

is not working in my system even if I use clean, run qmake, rebuild. I don't know how to correct this.

elieserdejesus:
@jonjamcam , I think the correct rule name is **NinjamTrackView**, not *BaseTrackView*. Possibly there is a rule in NinjamWindow.css overriding your changes. If you write two rules, one for BaseTrackView and another for NinjamTrackView the last will be applyed.

jonjamcam:
@elieserdejesus I found the problem .Just try removing the comments for that particular line  in the original code: 

`/* track when is not xmiting */`

I tested and now the code works. i will make a new PR.

PR #398, opened 2016-05-26
Flat Theme top panel color fix to avoid track layer mismatch illusion

jonjamcam:
This is the final result:

image


jonjamcam:
PR moved to https://github.com/elieserdejesus/JamTaba/pull/399

PR #399, opened 2016-05-26
Round theme ninjamer disabled track color fix

jonjamcam:
I found that the comment in this line is blocking the code, so this PR removes the comment.


elieserdejesus:
It's merged @jonjamcam , thanks!

PR #451, opened 2016-06-13
Move file handling to writers

pljones:
Now files are only created if the writer is enabled.

### DIFF
--- a/src/Common/recorder/ClipSortLogGenerator.h
+++ b/src/Common/recorder/ClipSortLogGenerator.h
@@ -17,11 +17,7 @@ public:
     {
         return QCoreApplication::translate("Recorder::ClipSortLogGenerator", "Generate Reaper 'clipsort.log' file");
     }
-    inline void setJamDir(QString newJamName, QString recordBasePath) override {
-        QDir parentDir(QDir(recordBasePath).absoluteFilePath(newJamName));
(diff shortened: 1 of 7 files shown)


jonjamcam:
@pljones I just tested this branch and it is working ok now. No empty folders anymore. And if you check both options, you'll get the jam recorded twice, but this is up to the user if this is the intention. :+1:

src/Common/recorder/ClipSortLogGenerator.h
```diff
@@ -17,6 +17,14 @@ class ClipSortLogGenerator : public JamMetadataWriter
     {
         return QCoreApplication::translate("Recorder::ClipSortLogGenerator", "Generate Reaper 'clipsort.log' file");
     }
+    inline void setJamDir(QString newJamName, QString recordBasePath) override {
```
elieserdejesus: @pljones , a have just one little code review: too much code for a inline function here?

pljones: Whoops yes - I'd forgotten to make the subdir originally.  I'll move it.

pljones:
@elieserdejesus, I've updated per comments.

elieserdejesus:
thanks @pljones.

PR #453, opened 2016-06-14
[WIP] Creating exclusive jam folders

elieserdejesus:
Creating jam folders using localized date string (including the current second in system clock).

This solution is trusting in current clock secong to create a different jam folder for every started jam. If the user is fast enough to connect in 2 different servers in one second the solution will fail :)


jonjamcam:
>  If the user is fast enough to connect in 2 different servers in one second the solution will fail :)

:smile:

PR #454, opened 2016-06-14
little typing mistake in german translation

jonjamcam:
Just added the "t" to "cha" L. 739


elieserdejesus:
Thanks @jonjamcam , I think Magnus is not using QtLinguist :)

PR #460, opened 2016-06-18
Allowing custom vorbis encoder quality

elieserdejesus:
- Allowing custom vorbis encoder quality in .json config file
- Sanitizing json values in range [-0.1, 3]
- Adding GUI

### DIFF
--- a/src/Common/MainController.cpp
+++ b/src/Common/MainController.cpp
@@ -71,6 +71,13 @@ void MainController::setSampleRate(int newSampleRate)
     settings.setSampleRate(newSampleRate);
 }
 
+void MainController::setEncodingQuality(float newEncodingQuality)
+{
+    settings.setEncodingQuality(newEncodingQuality);
+    if (isPlayingInNinjamRoom())
+        ninjamController->recreateEncoders();
+}
+
 void MainController::finishUploads()
 {
     foreach (int channelIndex, intervalsToUpload.keys())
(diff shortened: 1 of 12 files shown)


jonjamcam:
It's working OK. values greater than 0.3 are not allowed, value 0 is working OK and values less than -0.1 are not allowed either.

elieserdejesus:
@jonjamcam , I added a GUI to choose the 3 quality levels. If an user manually change the json a 4th (Custom quality) will appear, just to have some feedback about the non pre defined value.

**Important**: this option is available for standalone only because the **audio** preferences tab is not available in the Vst Plugin. We can think in a better solution in future versions.

image

jonjamcam:
Tested. Working great! Very useful feature.

jonjamcam:
> his option is available for standalone only because the audio preferences tab is not available in the Vst Plugin

I was thinking about this and one idea came to mind: Create a **network** tab (F10). This can be useful if later we need more options related to the internet, LAN, etc.

elieserdejesus:
@jonjamcam , I think the encoding quality is related with both audio and network. But in my mind the weights are 70% for audio and 30% for network. I'm shure these **weigths** are very subjective. My opinion is based in the assumption (subjective too) Jamtaba users are more interested in audio quality than network usage. 

A simple solution is just hide the non used audio controls in **Audio Tab**, but I'm not shure the current layout will work. At moment the entire audio tab is hidden in VSt plugin.

PR #463, opened 2016-06-21
fr_FR L. 515 fix

jonjamcam:
I found this text was marked "unfinish" in this update, but if you go back to previous release, the translation is there.

image


elieserdejesus:
thanks @jonjamcam .

PR #465, opened 2016-06-21
ja_JP fix

jonjamcam:
I replaced this two "unfinish" items from the previous release.

image


elieserdejesus:
Thanks @jonjamcam !

Issue #477, opened 2016-06-28
Low cut button for ninjam tracks


elieserdejesus:
Add a low cut button in ninjam tracks to "clean the mix".

PR #478, opened 2016-06-28
Low cut

elieserdejesus:
Working in #477 (low cut button in ninjam tracks)


elieserdejesus:
@jonjamcam , I added a new cool and usefull feature:
image

jonjamcam:
@elieserdejesus nice. My tests work OK here. In theory I suspect the frecuency (220 Hz) could be too much for some instruments like Bass guitar, but I need to test while jamming to have a definite opinion. This is a good feature to have.

elieserdejesus:
@jonjamcam , apply a low cut in the bass or drums (I'm thinking in the kick)  it's always a bad idea. The low cut (high pass filter) is usefull for the other instruments in general, but is not recommended for "low freq" instruments. Changing the cut off frequency to 110 Hz will not solve the problem for bass or drums.

elieserdejesus:
@jonjamcam , I understand your point, but I think in this case (sounds not very good processed in low freqs) is better solve the situation using an common EQ (a band pass filter) and reduce the low freqs. Reduce the low freqs using an band pass is less agressive than **cut all (or almost) the low** freqs using a high pass.
The main application is cut low freqs in guitars and keyboards (very common in ninjam) to keep the drum kick and bass low freqs sounding clear.

jonjamcam:
>in this case (sounds not very good processed in low freqs) is better solve the situation using an common EQ (a band pass filter) and reduce the low freqs.

Yes, of course. I didn't thought you had that one planned.

> The main application is cut low freqs in guitars and keyboards (very common in ninjam) to keep the drum kick and bass low freqs sounding clear.

Sure that's pretty well covered.

PR #481, opened 2016-07-04
fix Japanese translation.

takeshitaleo:
re-pull request:)


elieserdejesus:
thanks @takeshitaleo , is perfect clean now :)

Issue #500, opened 2016-07-16
little problem with the "copy", "select all" functions and "right click menu" inside the chat widget


jonjamcam:
So let's say you want to copy the text of a chat message. You select the text and right click:

![image](https://cloud.githubusercontent.com/assets/15310433/16897330/047308d0-4b7c-11e6-9754-ad4e9b3fec3c.png)

1.- In this case, the is copied to the clipboard **only** if you select **copy** in the menu above, but it's not copied if you type **ctrl + c** 

To select all the text you have to use the right click menu or double click the text, but if you use **ctrl + a** it's not selecting the text at all. (these are only a problem with the Jamtaba VST plugin as stated at the bottom if this comment)

2.- Another thing I noticed is that if you position the cursor anywhere across the text, but do not select anything, the right click menu will appear, but won't work:

![image](https://cloud.githubusercontent.com/assets/15310433/16897306/34a47a44-4b7b-11e6-936a-0ca79ad43ef5.png)

 I mean the **copy**, **copy link** and **select all** options do not produce any result. and it looks like a bug (to me).

![image](https://cloud.githubusercontent.com/assets/15310433/16897322/ce5cddac-4b7b-11e6-8b24-f166c28c003b.png)

3.- The right click menu appears even if I point a blank space:

![image](https://cloud.githubusercontent.com/assets/15310433/16897400/1604aeb2-4b7e-11e6-9baa-60f7df64d1a2.png)

and it doesn't work as stated in **2**

EDIT: 1,2 and 3 were tested in Jamtaba VST plugin x64 inside Reaper x64. For the Standalone version of Jamtaba the only problems that apply is numbers **2**and **3**.

elieserdejesus:
@jonjamcam , **2 and 3** are issues, but I think the issue described in **1** is the default behavior in all software. I tested the same situation here in text editors (word and notepad) and have the same result (control + c not work):
1 - type a word in some text editor
2 - double click to select the word
3 - Right click to open the context menu (the Copy option is activated, we have something to "copy")
4 - Press CONTROL + C
5 - Windows emited a sound warning me this is not allowed. The text is not copyed to clipboard.

I think the problem in 2 and 3 are related, the options are appearing but there s nothing to copy or select. This menu behavior is default in Qt and unfortunally Qt is not handling these very common situations. I will investigate how improve these context menus.

jonjamcam:
[...]

PR #502, opened 2016-07-19
Fix chat message context menu

elieserdejesus:
Working in issue #500

The "non working" menu items are just a problem of stylesheet. The previous code was using same color for disabled and enabled menu items. Now the copy, paste, select all options appear with different colors when disabled.

Control + A and Control + C working for chat messages.


jonjamcam:
tested and working OK. I already close issue https://github.com/elieserdejesus/JamTaba/issues/500 but I forgot you need to merge first. oops!

elieserdejesus:
no problem @jonjamcam.

Issue #474, opened 2016-06-24
Automatic Recognition for new AUDIO device shows unsupported value for buffer size


jonjamcam:
It is needed to for **F5** to be pressed twice when plugging in new USB interface to get the correct buffer size value for the new device. To reproduce the problem:

1.- Launch Jamtaba with new interface **not** plugged in. Only ASIO4ALL appears in drivers list (in this case):

![image](https://cloud.githubusercontent.com/assets/15310433/16351001/cf26b96e-3a38-11e6-8f3e-fde3a392392e.png)

 Buffer size is the default value for Jamtaba (128). Note this value is not supported by all interfaces (i.e. ME-25 uses 96, 144, 192, 288, etc).

2.- Plug in USB interface, wait 5 seconds and press F5. The new interface (ME-25 in this case) appears, but with wrong value for buffer size (should be 288 in this case):

![image](https://cloud.githubusercontent.com/assets/15310433/16351171/21453b66-3a3a-11e6-86a3-93b450c201e1.png)

The correct values does not appear in the list:

![image](https://cloud.githubusercontent.com/assets/15310433/16351226/778c82ae-3a3a-11e6-84ba-3fcec6fd8bed.png)

 as the ASIO panel shows 288:

![image](https://cloud.githubusercontent.com/assets/15310433/16351294/f314c2ec-3a3a-11e6-9042-24f107eda710.png)

To get the correct values you need to press OK and F5 a second time:

![image](https://cloud.githubusercontent.com/assets/15310433/16351405/bc0b9c16-3a3b-11e6-8a92-5ef5b19738ea.png)

EDIT: A refresh button could be enough or an automatic refresh after using the ASIO panel button.

PR #505, opened 2016-07-19
Fixing wrong values in ASIO buffer size combobox

elieserdejesus:
fixing #474

### DIFF
--- a/src/Standalone/gui/PreferencesDialogStandalone.cpp
+++ b/src/Standalone/gui/PreferencesDialogStandalone.cpp
@@ -362,6 +362,7 @@ void StandalonePreferencesDialog::changeAudioDevice(int index)
     populateFirstInputCombo();
     populateFirstOutputCombo();
     populateSampleRateCombo();
+    populateBufferSizeCombo();
 }
 
 void StandalonePreferencesDialog::accept()


elieserdejesus:
@jonjamcam , I'm working in the issue #474 . Now when we change the audio driver the buffer size combobox is updated with supported values. Can you test your devices, please?

jonjamcam:
@elieserdejesus Yes, the list is correct as soon as you select the driver:

image

elieserdejesus:
@jonjamcam , just to be shure, can you test the scenario described in the issue #474 ? You reported a initial buffer size of 128 using ASIO4ALL and change to ME-25 after some time. 

What you see when you change from **ASIO4ALL + 128 samples** to ME-25?

jonjamcam:
>What you see when you change from ASIO4ALL + 128 samples to ME-25?

That was actually the test. My computer always uses ASIO4ALL as default driver. So I opened the preferences and and changed to ME-25. The buffers and sample rate changed instantly from 48000/64 to 44100/96

elieserdejesus:
@jonjamcam , I see the values 64 and 96 are the minimum buffer size values for your 2 devices. In this case, I'm worried Jamtaba is just keeping the first item in the combobox.

can you do a last test **ME-25 + 6144 samples** to ASIO4ALL? What value is selected in ASIO4ALL in this case?

I can't do these tests because my ASIO4ALL and my Fast track ultra are showing exactly same buffer size values list.

jonjamcam:
>My computer always uses ASIO4ALL as default driver

@elieserdejesus First let me be more specific on this issue. If the ME-25 is not connected, then ASIO4ALL is the default driver, but if I start Jamtaba with the ME-25 plugged in and turned ON, then the ME-25 acts as default. So the question is **out of the question** :)

>can you do a last test ME-25 + 6144 samples to ASIO4ALL? What value is selected in ASIO4ALL in this case?

Ok. So we start with these values:

image

The buffer value of 6144 is kept even after restart. Now we change driver to ASIO4ALL:

image

The buffer size and Sample Rate shown by default are the minimum (64/44100). Wchange that to other values (2048/48000)

image

Now back to ME-25 again values are the minimum, not the ones I chose before:

image

elieserdejesus:
@jonjamcam, 
> Now back to ME-25 again values are the minimum, not the ones I chose before

Jamtaba is not remembering the "preferred" buffer size for each ASIO driver, but I think this is a good idea for a new feature. Another possibility is use the ASIO preferred buffer size. If I remember correctly asio drivers can report the preferred buffer size. But I think this is another feature/issue.

I think this issue is solved, right? no more problems with invalid buffer size values.

jonjamcam:
> I think this issue is solved, right?

Yes. It's solved.

Issue #430, opened 2016-06-05
ASIO4ALL overload when using LISTEN button in Jamtaba standalone X64


jonjamcam:
![image](https://cloud.githubusercontent.com/assets/15310433/15807560/cf7b8580-2b37-11e6-8368-4f42522bbcd3.png)

This is NOT happening with the vst plugin (x64) and standalone 32bits. This started to happen for me in 2.0.14. I first thought there was some problem with my system, but I installed old 2.0.13 and I don't have any problem with the x64 standalone and the listen button.

jonjamcam:
Testing in 2.0.15 pre release now. I suspect this is an ASIO4ALL problem. Tweaking some settings in the ASIO4ALL panel got it workong fine, but after close and relaunch, the problem starts again.

 Using ASIO4ALL v2.13 on WIN7 64bit

elieserdejesus:
@jonjamcam , I'm using Asio4ALL 2.13 here too, no problems using 128 as buffer size.

But I'm suspecting mp3 decoder is blocking the audio thread.

jonjamcam:
>I'm suspecting mp3 decoder is blocking the audio thread.

One question:

is it possible for this to happen ONLY with the 64bit standalone? I use the 64bit VST plugin and the Standalone 32bit with no problems.

I tested 2.0.14 in another laptop with ASIO4ALL 2.13 too and I have the a similar problem: The 32bit standalone version works pretty good there (with buffer 128), but the 64bit version does not work at all (maybe some ASIO4ALL/Audio hardware problem)

I will do more experiments. It's not a big deal for me because the 32bit version works like a charm.

elieserdejesus:
@jonjamcam , I'm using the 64 bits version here (both standalone and plugin) without problems.

elieserdejesus:
@jonjamcam , this problem is a bit complicated to fix, probabily it's not a quick fix. Since I can't reproduce the bug in my JT 64 bits I will release v2.0.15. If this is really a bug I can work in this task in the next round.

jonjamcam:
Sure, no hurries. I'm good with the 32bit version.

jonjamcam:
I think I found the problem (I hope so). I did some tests using my ME-25 pedalboard as ASIO interface in two different laptops and ASIO4ALL too for comparison.

[...]

PR #514, opened 2016-08-23
Fixing ASIO4ALL overload when listening public rooms

elieserdejesus:
- Decoding small mp3 packets to avoid blocking the audio thread
- Refactoring some bad named signal/slots

### DIFF
--- a/src/Common/MainController.cpp
+++ b/src/Common/MainController.cpp
@@ -704,7 +704,7 @@ void MainController::start()
             setTheme(settings.getTheme());
 
         qCInfo(jtCore) << "Creating roomStreamer ...";
-        roomStreamer.reset(new Audio::NinjamRoomStreamerNode()); // new Audio::AudioFileStreamerNode(":/teste.mp3");
+        roomStreamer.reset(new Audio::PublicRoomStreamerNode()); // new Audio::AudioFileStreamerNode(":/teste.mp3");
         this->audioMixer.addNode(roomStreamer.data());
 
         QObject::connect(&ninjamService, SIGNAL(connectedInServer(const Ninjam::Server &)), this,
--- a/src/Common/MainController.cpp
+++ b/src/Common/MainController.cpp
@@ -32,6 +32,8 @@ MainController::MainController(const Settings &settings) :
     usersDataCache(Configurator::getInstance()->getCacheDir())
 {
 
+    qRegisterMetaType<Audio::SamplesBuffer>("Audio::SamplesBuffer");
+
     QDir cacheDir = Configurator::getInstance()->getCacheDir();
     ipToLocationResolver.reset( new Geo::WebIpToLocationResolver(cacheDir));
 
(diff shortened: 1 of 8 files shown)


elieserdejesus:
@jonjamcam , I can't reproduce the bug described in #430 in my machine, but I'm suspecting the mp3 decoder. I changed the code to decode small mp3 packets and avoid freeze the audio thread (ASIO4ALL) while decoding. In my tests it's working ok with 32/64 bits and with small and large audio buffer sizes (128 - 2048).

Can you test this branch in your machine to see if ASI04ALL is overloaded?

jonjamcam:
@elieserdejesus I can only compile x32 in my system.

jonjamcam:
@elieserdejesus Tested, but I still have the problem. I checked 2.0.13 and I can confirm it was the last version that was working OK. If you need I can do a video so you can see the problem better.

elieserdejesus:
@jonjamcam 
> but I still have the problem. 

The 2 problems? Audio crackles and red ASIO4ALL icon?

jonjamcam:
>The 2 problems?

Just the ASIO4ALL icon problem. No crackles in my tests so far.

elieserdejesus:
@jonjamcam , can you test again? I'm using another approach to avoid blocking the audio/ASIO4ALL thread.

[Jamtaba2.zip](https://github.com/elieserdejesus/JamTaba/files/434723/Jamtaba2.zip)

jonjamcam:
@elieserdejesus This last version solves the overload problem, but the audio of the room skips (like a scratched CD) and after a while jamtaba crashes.

elieserdejesus:
@jonjamcam , can you test/compile this branch again? I added some log messages to try understand the problem.

jonjamcam:
@elieserdejesus Works pretty good now. No more skipped sound and no crash. This are the messages:

image

elieserdejesus:
Nice news @jonjamcam , I tested in Linux and is working too. I'm attaching the x64 version:
[Jamtaba2.zip](https://github.com/elieserdejesus/JamTaba/files/437385/Jamtaba2.zip)

jonjamcam:
@elieserdejesus x64 working OK! Good Job!

elieserdejesus:
Thanks @jonjamcam !

jonjamcam:
@elieserdejesus I tested a compiled version (32bit standalone) in my old Pentium 4 XP machine (1 processor) and the listen function almost freezes the PC. I don't know if anyone is using XP these days, My question this PR will work on multiprocessor computers only?

elieserdejesus:
@jonjamcam , the current solution works better in multiprocessor computers, but this is not mandatory. At moment Jamtaba is using different threads to play audio (portaudio), decode mp3 and GUI (Qt thread). Qt use another threads internally to run network code too. 

In multiprocessor computers these threads can run in different cores, and the result is very good from a processor usage point of view. In a mono core computer all threads will run/compete for the only processor. These threads will be scheduled in the processor, every thread using a little time slice.

So, in theory these multi thread solutions will work in mono core processors, but the result can be not so good :(

jonjamcam:
I see. I like the thread separation. My win7 Laptop is rather old (2 cores), but Jamtaba works like a Charm, specially now.

 Anyway it's only the listen function that has problems. And I suspect XP is not longer used except for old software compatibility, so not big deal.

PR #563, opened 2016-09-04
Update en.ts

jonjamcam:


elieserdejesus:
Thanks @jonjamcam , I will compile a new pre-release and upload in some minutes.

PR #567, opened 2016-09-06
improving and matching colors for all themes (comboboxes and selections)

jonjamcam:
1.- I removed the selection color in the common file and matched themes individually

2.- Some improving and matching colors for the ninjam panel (the idea here is to dim a little the labels and bright the combo box information):

image

image

image

image

3.- fixed squared buttons when hover for the volcano theme


elieserdejesus:
nice @jonjamcam !

PR #584, opened 2016-09-11
[WIP] Loading theme from external folder

elieserdejesus:
- [x] Export default themes from resources to Jamtaba install folder
- [x] Load themes from external folder 
- [x] Improve the behavior when a empty theme dir is loaded
- [x] Implement night mode suffix in theme names
- [x] Clean unecessary aliases in resources file


elieserdejesus:
@jonjamcam , now the themes are loaded from external directory. The default themes are copyed from resources folder to AppData/Jamtaba 2/Themes. 

Now is possible change a CSS file and see the result just executing Jamtaba, no need to recompile.

jonjamcam:
Yeah, I tested it. Very easy now. The only thing you can't control is the "night mode" for the map.

elieserdejesus:
Exactly @jonjamcam , at moment the *night mode* is hardcoded in **MainWindow.cpp**, line **119**.

A simple idea to solve this problem is add a prefix or suffix in the theme folder name to denote the "night mode". For example:
```
Themes
   |_ Black 
   |_ Black_nm       //the suffix '_nm' can be used to denote a "night mode" theme
   |_ Volcano_nm
```
This suffix can be striped when building the themes menu.

PR #586, opened 2016-09-12
Finishing preset delete

elieserdejesus:


jonjamcam:
Delete preset working OK here.

PR #667, opened 2017-01-16
Windows Performance Monitor

jonjamcam:


elieserdejesus:
@jonjamcam , this is the result in my machine, Win 10, 4 cores, JamTaba running in debug mode (slow, more CPU usage) using a loop in SampleTank 3 and processed with Amplitube, just to use some CPU.
image

jonjamcam:
I see the difference. The author of the code (Jeremy Friesner) wrote:

"_(*) Okay, you might get a bit more resolution on a multicore system; e.g. if you measured instantaneous CPU usage on a quad-core CPU you might find that at that particular instant in time, three cores were idle and one core was active, and call that 25% load... and of course there are things like Intel's SpeedStep that actually varies the CPU's clock rate as a way to manage power consumption; but we'll ignore those complications for the time being :)_"

Do you have the same result when all cores are used in a similar way? In my tests normally the load is more or less the same throughout all cores:



and that gives a more consistent result with the task manager value.

jonjamcam:
Because of the accuracy problem in the CPU meter,  what do you think @elieserdejesus  about keeping the memory meter only for now?

image

elieserdejesus:
@jonjamcam , I think keep memory usage only it's a good idea. 

I see you commented the old code in PerformanceMonitor. Please try rollback the old code to avoid introduce non used changes.

jonjamcam:
> Please try rollback the old code to avoid introduce non used changes.

OK. I'll keep the old code just adding the modifications for the memory meter

elieserdejesus:
@jonjamcam , Can I merge this PR?

jonjamcam:
> Can I merge this PR?

Sure. Thank you.

elieserdejesus:
merged, thanks @jonjamcam !

Issue #659, opened 2017-01-02
json and cache configuration files are reset in 2.0.19 as expected


jonjamcam:
Due to directory change from "_\AppData\Local\Jamtaba 2_" to "_C\AppData\Roaming\Jamtaba 2_" all cache and json configuration is "lost" for average user. If the user tweaked the themes, that is also lost.

A solution may be to have the installer move all these files automatically to the new path (if old files are detected) or if this is not possible, show a notification for the user to move them manually.

PR #675, opened 2017-01-23
Installer copying old appData/Local folder to appData/Roaming 

elieserdejesus:
solving issue #659

- Since v 2.0.19 the data folder was changed from 'appData/Local/Jamtaba 2' to 'appData/Roaming/JamTaba 2' folder. The installer is checking if the 'appData/Local/Jamtaba 2' exists in user machine and copying files recursivelly to Roaming folder.

- Uninstall script updated to remove files from Roaming folder too.

### DIFF
--- a/installers/windows/Installer script common.iss
+++ b/installers/windows/Installer script common.iss
@@ -57,6 +57,7 @@ Source: {#BuildDir}\Standalone\release\Jamtaba2.exe; DestDir: {app}; Flags: igno
 Source: {#BuildDir}\Standalone\release\VstScanner.exe; DestDir: {app}; Flags: ignoreversion replacesameversion
 Source: {#BuildDir}\VstPlugin\release\JamtabaVST2.dll; DestDir: {code:GetVST2Dir}; Flags: ignoreversion replacesameversion
 Source: ..\..\PROJECTS\Jamtaba2.ico; DestDir: {app}; Flags: ignoreversion
+Source: {localappdata}/Jamtaba 2/*.*; DestDir: {userappdata}/JamTaba 2; Flags: external recursesubdirs skipifsourcedoesntexist
 
 [Icons]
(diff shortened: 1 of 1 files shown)


elieserdejesus:
@jonjamcam , the installer is copying user data from **appData/Local/Jamtaba 2** to *appData/Roaming/JamTaba 2*. Can you test in your machine? 

[Jamtaba 2 (64 bits) installer.zip](https://github.com/elieserdejesus/JamTaba/files/724001/Jamtaba.2.64.bits.installer.zip)

jonjamcam:
> the installer is copying user data from appData/Local/Jamtaba 2 to appData/Roaming/JamTaba 2

Yeah, working fine here :D

elieserdejesus:
nice, thanks @jonjamcam .

jonjamcam:
I tested in win10 too where I already had the _roaming/Jamtaba 2_ folder. In this case when trying to replace **logging.ini** I had to press _omit_ because the file was read only.

elieserdejesus:
> press omit

@jonjamcam , this is a windows dialog with a "omit" button? I can't understand, I tested in win 10 here is ok.

**I noticed another problem**:  the file **logging.ini** is not exported in my machine. Can you confirm?

jonjamcam:
>  is a windows dialog with a "omit" button

Well, it's my translation as it appears here in spanish, similar to this:



. When the installer tries to overwrite the file a dialog appears (in spanish) where you can choose.

jonjamcam:
I installed again. This is the problem:

image

The options are: Cancel, Try again, omit.

Try again doesn't work, cancel will cacel installation, so omit works, but do not copy the file.

elieserdejesus:
I reproduced the problem here. Fixed in https://github.com/elieserdejesus/JamTaba/commit/4f0290d9c86b339e64d70b6a3f32435f57a6265f

Issue #676, opened 2017-01-23
Full view mode JamRoomWindow often appears as in minimode


jonjamcam:
This happens:

![image](https://cloud.githubusercontent.com/assets/15310433/22212592/06f3ceac-e170-11e6-8622-e9625432a179.png)

and returns to normal automatically after server refresh:

![image](https://cloud.githubusercontent.com/assets/15310433/22212627/2d52f852-e170-11e6-9468-2a788dcab114.png)

Issue #669, opened 2017-01-18
Some ideas for Local Tracks enhancement


jonjamcam:
I noticed a smaller size for the Full View Mode starting in 2.0.19:

![image](https://cloud.githubusercontent.com/assets/15310433/22047291/dd6eeaa4-dd04-11e6-924a-f55fc926ada7.png)

 This new size is great, specially for the vst plugin, but now the "Full View layout" will have very little space for the level meters:

![image](https://cloud.githubusercontent.com/assets/15310433/22047354/4e57a116-dd05-11e6-90f1-deacfcb21b37.png)

To solve this problem we could make the "Mini Mode Layout" also available in the Full View Mode, specifically for the vst plugin as it's not resizable. The standalone can be maximized, so no issue there.

![image](https://cloud.githubusercontent.com/assets/15310433/22047575/a7243ad8-dd06-11e6-9905-458b7c568d24.png)

Another problem is we'll have now less space for remote tracks (specially when the room is full), so we can save a lot of space if a Wide/Narrow button is available for Local Tracks (both vst and standalone).

 There's some unused space here to put a little button for example:

![image](https://cloud.githubusercontent.com/assets/15310433/22047421/ba3063e6-dd05-11e6-85d9-0b6be11a4d34.png)

***
Related, but off topic, @elieserdejesus what do you think about having a third view option "_Big screen Mode_" or "_Extended View Mode_"for the vst plugin with resolution in the range of 1280x720, 1360x790 or maybe even 1920 × 1080?

I'm thinking about Big LCD TV resolutions. I can test here in two big screen TV's we have at home.

elieserdejesus:
@jonjamcam 
> what do you think about having a third view option

I'm in the opposite direction :)  I was thinking in remove these "Mini and Full" view options. I think these 2 options are confusing, no other software use these options and users are not used. The Full Screen option is clear and other software are using, but Mini and Full View are not clear enough in my opition. If you maximize JT window and switch from Full to Mini View you will see the effect is minimum. 

Some ideas:
1 - Drop Mini mode
2 - put the meters along side with level fader
3- VST/AU PLugin window start with small size (current mini mode size), but will be resizable

jonjamcam:
> Some ideas:
1 - Drop Mini mode
2 - put the meters along side with level fader
3- VST/AU PLugin window start with small size (current mini mode size), but will be resizable

Yeah. Much better. The resizeable vst is the best solution.

[...]

PR #677, opened 2017-01-24
[WIP] Dropping minimode

elieserdejesus:
solving  #676, #669

- [x] Dropping mini mode
- [x] Adding window resize submenu in View menu
- [x] Storing and remembering window size (both standalone and VST)

- [x] The Vst plugin keeps the square (4:3 ?) aspect ratio when resized. It could be better if the aspect ratio was a little more rectangle shaped, like on the standalone

- [x] Local tracks are small all the time in the vst plugin version, but I noticed the standalone resize them to a good size when maximized

- [x] When you connect to a server the "big local tracks" are reduced again, and they keep the small size until restart


elieserdejesus:
@jonjamcam , can you help testing this PR? I dropped minimode and added menu entries to resize the VST window.

[Jamtaba 2 (64 bits) installer.zip](https://github.com/elieserdejesus/JamTaba/files/728009/Jamtaba.2.64.bits.installer.zip)

jonjamcam:
My tests show:

1. The Vst plugin keeps the square (4:3 ?) aspect ratio when resized. This is a snap from a 1920x1080 TV.:

image

 It could be better if the aspect ratio was a little more _rectangle shaped_, like on the standalone version (if possible):

image

2.- The local tracks are small all the time in the vst plugin version, but I noticed the standalone resize them to a good size when maximized. If the plugin could have the same feature when the size is larger than 800x600 (or any other resolution that's not considered small) it would be better for larger screens:

image

 Overall the result is very good. Big improvement, specially for big screens.

elieserdejesus:
@jonjamcam , I fixed the issues you commented before. It's much better now. Improved aspect ratio in plugin window, the resize menu entries are disabled when is not possible change the window size, the 2 collumn layout (public servers) is used always horizontal space is enough, the local tracks are always wide, except in plugins (VST/AU) when minimum window size is used and there are 2 local channels. 

Can you run a final inspection, please?

[Jamtaba 2 (64 bits) installer.zip](https://github.com/elieserdejesus/JamTaba/files/731195/Jamtaba.2.64.bits.installer.zip)

jonjamcam:
> Can you run a final inspection, please?

Works great. I tested in the TV and the maximun size is perfect. Local tracks are correct for each size now.

Issue #102, opened 2015-11-07
Input audio and midi at the same time


jonjamcam:
Some plugins/instruments need you to input audio and midi at the same time (vocoders, loopers). Currently there's no way to do that in Jamtaba.

Reaper solves this issue by flexible routing capabilities. So you create a new track in Reaper and send it to where you need.

I was thinking about this lately and there's another big problem. What if you have two or three keyboards and need them all to connect to KONTAKT for example?

Let's try to think in simple terms. You have already a rack to choose plugins with four racks. I use at most two and I need two extra inputs. My question is could this racks be used as extra MIDI inputs when not used by plugins?

Midi has this structure separated in channels, so mixing midi connections does not mean mixing midi information if they are on separate channels. Audio is a different matter. you always need a mixer.

So in short.......same configuration as now, but the first two racks can be used as midi inputs or vstplugins, but not both. what do you think?

elieserdejesus:
@jonjamcam , at moment we have *no input*, *audio*,  and *midi* options in track input panel. I was thinking in add a new option *audio + MIDI*. Seens to me this solve the vocoder issue. What you think about?

But you mentioned the **extra MIDI inputs** too. but seens to me this is a sepparated issue?

jonjamcam:
> add a new option audio + MIDI.

Yeah, it's good enough.

 (Now that I think about it what I wrote previously (https://github.com/elieserdejesus/JamTaba/issues/102#issue-115690497) was too complicated. It's good to think in simpler terms)

elieserdejesus:
@jonjamcam , I realized this feature is complicated from the UI perspective. The problem is how users will select audio and MIDI input? 

I think in some options and at moment the best solution I can imagine is allow audio and midi everytime. So users always will see a midi input selection and an audio input selection. User can select "no input" to both audio and midi, can select valid audio input or valid midi input, or both. 

I think this idea work, but I don't it's good from an end user perspective. The UI will be more complex, and just to support a very exceptional case: AUDIO + MIDI in vocoders, etc. 

I think we need think more about this feature until we have a good solution.

Please tell me if you have another ideas.

jonjamcam:
> the UI will be more complex, and just to support a very exceptional case: AUDIO + MIDI in vocoders, etc.
> Please tell me if you have another ideas.

Sure. No problem. For the moment I have this one idea:

 Use the subchannel as the additional input (midi or audio) adding a button or a checkbox, for example.

[...]

PR #682, opened 2017-01-28
[WIP] Allowing midi controlled audio tracks (like vocoders)

elieserdejesus:
- [x] Add a checkbox in midi tools dialog to de/activate midi input routing.
- [x] Implementing midi routing in LocalInputNode
- [x] Implementing visual feedback in local tracks to show when midi routing is activated.
- [x] Disable buttons in second track when midi routing is enabled
- [x] Midi route checkbox will be available only in second subchannel.
- [x] Remember/save midi route status
- [x] Save/load midi route flag in presets
- [x] Disable midi input in second track when midi routing is enabled
- [x] Fix midi route drawing when local tracks are collapsed
- [x] disable all audio related controls (pan, fader, plugins etc) when routing midi.
- [x] If the user changes his mind and decides to use a normal subchannel (selecting audio input) the midi routing should be disabled
- [x] Fix midi activity meter when routing midi
- [x] Fix not intentional changes in Black theme
- [x] Reset track do not enable previously disabled buttons.

### DIFF
--- a/src/Common/MainController.cpp
+++ b/src/Common/MainController.cpp
@@ -263,7 +263,7 @@ void MainController::removeInputTrackNode(int inputTrackIndex)
     if (inputTracks.contains(inputTrackIndex)) {
         // remove from group
         Audio::LocalInputNode *inputTrack = inputTracks[inputTrackIndex];
-        int trackGroupIndex = inputTrack->getGroupChannelIndex();
+        int trackGroupIndex = inputTrack->getChanneGrouplIndex();
         if (trackGroups.contains(trackGroupIndex)) {
             trackGroups[trackGroupIndex]->removeInput(inputTrack);
             if (trackGroups[trackGroupIndex]->isEmpty())
@@ -281,12 +281,12 @@ int MainController::addInputTrackNode(Audio::LocalInputNode *inputTrackNode)
     inputTracks.insert(inputTrackID, inputTrackNode);
     addTrack(inputTrackID, inputTrackNode);
 
-    int trackGroupIndex = inputTrackNode->getGroupChannelIndex();
+    int trackGroupIndex = inputTrackNode->getChanneGrouplIndex();
     if (!trackGroups.contains(trackGroupIndex))
         trackGroups.insert(trackGroupIndex,
                            new Audio::LocalInputGroup(trackGroupIndex, inputTrackNode));
     else
-        trackGroups[trackGroupIndex]->addInput(inputTrackNode);
+        trackGroups[trackGroupIndex]->addInputNode(inputTrackNode);
 
     return inputTrackID;
 }
@@ -833,4 +833,13 @@ void MainController::storeMeteringSettings(bool showingMaxPeaks, quint8 meterOpt
 {
     settings.storeMeterOption(meterOption);
     settings.storeMeterShowingMaxPeaks(showingMaxPeaks);
+}
+
+Audio::LocalInputNode *MainController::getInputTrackInGroup(quint8 groupIndex, quint8 trackIndex) const
+{
+    Audio::LocalInputGroup *trackGroup = trackGroups[groupIndex];
+    if (!trackGroup)
+        return nullptr;
+
+    return trackGroup->getInputNode(trackIndex);
 }
--- a/src/Common/MainController.cpp
+++ b/src/Common/MainController.cpp
@@ -257,6 +257,13 @@ void MainController::saveEncodedAudio(const QString &userName, quint8 channelInd
 
 // +++++++++++++++++++++++++++++++++++++++++++++++
 
+void MainController::removeAllInputTracks()
+{
+    for(int trackID : inputTracks.keys()) {
+        removeInputTrackNode(trackID);
+    }
+}
+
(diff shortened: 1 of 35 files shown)


elieserdejesus:
@jonjamcam , I have a first version of issue #102.  The midi routing is working, tested using TAL vocoder.

image

jonjamcam:
Nice! it's working as expected. Tested with NI Vokator and mobius2.

jonjamcam:
@elieserdejesus just a reminder: to disable the feature/checkbox when the second subchannel changes to audio input/no input.

elieserdejesus:
> to disable the feature/checkbox when the second subchannel changes to audio input/no input.

@jonjamcam , I'm not shure if I understand the idea. When Midi tools dialog is opened is not possible change the input type to "no input" or "audio". If user select "no input" or "audio" the midi tools button is not visible, so is not possible open the midi dialog.  

 I'm missing something?

jonjamcam:
> I'm missing something?

I noticed this happening when testing the PR. I posted the comment in case you didn't notice it:

1.- add second subchannel
2.- click "route all audio to main..."
3.- Now buttons are disabled. OK so far.
4.- change input to audio in the second subchannel (simulating a change of mind of the user).
5.- buttons are still disabled.

So if the user changes his mind nad decides to use a normal subchannel (selecting audio input) the previous feature should be disabled.

elieserdejesus:
> So if the user changes his mind nad decides to use a normal subchannel (selecting audio input) the previous feature should be disabled.

Shure, adding this task in TODO list.

elieserdejesus:
@jonjamcam , I finished all tasks in this PR. Can you test before the merge?

jonjamcam:
works great! I just found some few GUI details. Functioning is perfect.

1.- **Stereo inverse button** for the second subchannel when routing is not disabled.
2.- Black theme did some modifications here https://github.com/elieserdejesus/JamTaba/pull/682/commits/0c0fa21e1d72261bb7f61fb9512c89c29e3356b9 to the stylesheet that were maybe not intentional? Audiometer is removed and some tabs hover code added.
3.- **reset track** do not enable previously disabled buttons (when routing was selected before) although is unchecks correctly the checkbox inside miditolls dialog.

elieserdejesus:
@jonjamcam , thanks for this detailed review.

you are right, some CSS changes was not intentional. I will fix this.

> Stereo inverse button for the second subchannel when routing selected is not disabled.

I can't reproduce this problem, I tested here is working ok.

jonjamcam:
> Stereo inverse button for the second subchannel when routing selected is not disabled.
> I can't reproduce this problem, I tested here is working ok. Can you check again, please?

I used _clean_, _run qmake_, _rebuild_ just to be sure and you are correct. The button is disabled as it should be. No problem here :D

jonjamcam:
> Stereo inverse button for the second subchannel when routing selected is not disabled.

@elieserdejesus I found the problem and how to reproduce it:

1.- add subchannel and activate midi routing checkbox.
2.- close Jamtaba.
3.- relaunch again.
4.- now the stereo inverse button is not disabled.

elieserdejesus:
@jonjamcam , it's fixed in https://github.com/elieserdejesus/JamTaba/commit/1e4ce31aba43bd38926a408b51357be34607773a

jonjamcam:
One other question. why there's a solid line arrow:

image

 arrow and the segmented one?

image

I see you use a segmented line when the first subchannel uses midi input, but in this case this midi input is disabled.

elieserdejesus:
@jonjamcam , my first idea was show a dashed line when the midi routing "path" is not complete, because 2nd channel is ok (input is midi and routing enabled) but the first channel is not an audio track. I was supposing only MIDI in 2dn channel and audio in first. If user select some different from audio in first channel the midi routing is "half" setted. 

I agree this is not clear enough. You mentioned 3 options, I think the first (allow dual midi input) is more clear for end users. 

My vote is drop the dashed line and allow dual midi input in first track. And you?

jonjamcam:
> My vote is drop the dashed line and allow dual midi input in first track. And you?

I agree with that option.

elieserdejesus:
@jonjamcam , I'm in a dilemma :)

If the idea is allow midi routing facilities I think make sense **allow routing when the first channel is setted to "no input" too**. *This means midi routing will be always possible*.
  
For example, you can activate the MIDI routing in 2nd channel and select "no input" in first channel. No sound will be produced, but the midi activity will show something in 2nd channel. Now you insert a **VSTi** in first subchannel and this plugin will receive midi input from 2nd subchannel. Off course this is not the easy way to setup a VSTi, but I think this is the expected result for experienced DAWs users. Break this expectation in general is not a good thing. If midi routing is disabled  when first track is "no input" users will need learn the specific situations where routing is allowed. If we can do something with **ZERO learning** is better :)

what do you think?

jonjamcam:
> If we can do something with ZERO learning is better :)

Yes I agree. Just allow midi routing to the first subchannel. The user decides what to input there. The simpler the beter.

elieserdejesus:
@jonjamcam , the new implementation is in PR #690.  Can you test, please?

PR #689, opened 2017-01-30
ALL THEMES fix for disabled Stereo inverse button

jonjamcam:
Volcano theme is the one that shows the "wrong effect" most notably, but this happens to the other themes as well. When restarting Jamtaba if there was any stereo inverse button **disabled** previously (because of a mono track or midi routing) the color of the button appears _grayish_:

image

Can you check if this happens in your system too @elieserdejesus , please?

The issue is solved deleting this line in the code, but I don't know if it being there has another purpose that I'm unaware of.

This is the result:

image

### DIFF
--- a/src/resources/css/common/BaseTrack.css
+++ b/src/resources/css/common/BaseTrack.css
@@ -39,8 +39,7 @@ QPushButton#buttonStereoInversion
     background-color: rgb(235, 235, 235);
 }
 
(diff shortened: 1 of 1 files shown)


elieserdejesus:
> The issue is solved deleting this line in the code, but I don't know if it being there has another purpose that I'm unaware of.

No problem @jonjamcam , the result is nice.

PR #690, opened 2017-01-31
[WIP] Improve midi routing

elieserdejesus:
- [x] Dropping dashed line in midi routing painting
- [x] First channel always receiving midi routing data
- [x] Avoiding duplicated midi notes (consumed midi messages are removed)
- [x] Dropping MidiMessageBuffer class (using std::vector<MidiMessage>)
- [x] Fix midi routing in first channel: 1) with midi controller connected setup the midi routing. 2) close Jamtaba and disconnect the midi controller; 3) open Jamtaba and the first subchannel will show the midi routing icon in top right.
- [x] Fix issue described in https://github.com/elieserdejesus/JamTaba/pull/690#issuecomment-276527398
- [x] Fix the "listen" issue described by Jon


jonjamcam:
@elieserdejesus It's not compiling in mingw32

image

elieserdejesus:
@jonjamcam , can you try again with the last commit?

jonjamcam:
Working now @elieserdejesus . So far all working as expected.

 Tested same midi information and channel sent to both inputs at the same time and there are no duplicates. Looks the first subchannel in the case takes all the input information. I really don't know if this kind of routing is of any practical use, but it's taken care of nicely. 

Sending two different midi input signals using the same and different channel is processed correctly and the midi meter is very useful in this case so you know which signal is going in and where.

Audio + midi working OK as before.

No input in subchannel 1 working as expected.

I'll try to connect tonight so I can test the full extent as I did before inside a server with real players. Any problems I'll report.

elieserdejesus:
thanks @jonjamcam 

> Tested same midi information and channel sent to both inputs at the same time and there are no duplicates. **Looks the first subchannel in the case takes all the input information**. 

Exactly, the first channel will "eat" the midi message.

jonjamcam:
@elieserdejesus I found one litttle problem when NOT using the midi routing. LEt's say you have two subchannels. If you load two different plugins in each subchannel and set both inputs to same input and channel, the second subchannel will not receive any midi data. I think this is OK happening only when using the routing function, but not when the subchannel is "normal"

jonjamcam:
>one litttle problem when NOT using the midi routing.
>
>Let's say you have two subchannels. If you load two different plugins in each subchannel and set both inputs to same input and channel, the second subchannel will not receive any midi data.

@elieserdejesus I was thinking about a simple soltion for the problem. If the midi messages are consumed, so duplicates are not allowed, is it possible to just copy the midi data when the same input and channel is selected in another channel or subchannel?

This, of course, only applies when NOT using midi routing.

elieserdejesus:
> is it possible to just copy the midi data when the same input and midi channel is selected in another jamtaba's channel or subchannel?

Yes @jonjamcam , I think is possible **always** send a copy of midi messages for each channel. Subchannels will "eat" the midi messages, but channels will receive a cloned (and full) midi messages pack. I will try this approach soon, I hope this will not generate side effects.

jonjamcam:
> I hope this will not generate side effects.

@elieserdejesus it was just a suggestion. If it's not a good idea we can try other approaches.

elieserdejesus:
> it was just a suggestion. If it's not a good idea we can try other approaches.

@jonjamcam , actually it's a very good idea, please don't get me wrong. I mentioned "side effects" because this feature is very deep in audio and MIDI engine. It's easy forget a detail, and as you can see every time I fix one thing another 2 appears :)

jonjamcam:
> every time I fix one thing another 2 appears :)

haha. true. I've been using Jamtaba for quite some time now and I know very well how it's supposed to work and I think I use 90% of all the features it has. I can detect very little differences and bugs easily. The good thing is I not only have fun using it to jam; I have a lot of fun testing it too :D

elieserdejesus:
Hi @jonjamcam , I finished the tasks in this PR. Can you do a new test in midi routing feature?

jonjamcam:
@elieserdejesus . It's all fixed now. Working perfect.

elieserdejesus:
thanks @jonjamcam .

Issue #668, opened 2017-01-17
[AU plugin] - keyboard press issues


elieserdejesus:
- In Reaper ENTER key is not working in chat. It's ok in Logic 9.
- Same problem with SPACE BAR.

elieserdejesus:
related with #657

elieserdejesus:
interesting link: [http://stackoverflow.com/questions/4752427/osx-detect-system-wide-keydown-events](http://stackoverflow.com/questions/4752427/osx-detect-system-wide-keydown-events)

Issue #661, opened 2017-01-09
Some problems in Cubase


elieserdejesus:
Reported by Andy:
- [ ] space sometimes stops and restarts cubase to then it is out of sync. 

- [x] when at 100bpm I have to set cubase to 101bpm then jamtasba see's the sync then i can set it back to 100 and start the sync.

Issue #657, opened 2016-12-31
[OSX AU] Return Key not working in chat


vector21:
Enter Key (Return Key) is not working as expected. At least on my Macbook (big return key over right shift key).

Works only when you hit "CMD + Enter" ...

Is this intentional? That might confuse users!! Like it confused me! :-)

But anyway... Very nice job on the AU port!!! Everything else works pretty much like as in the PC version.

elieserdejesus:
@vector21 , thanks for helping testing the AU plugin.

You tested in Reaper only?

What is your OSX version? Mavericks? Yosemite? El Captain?

vector21:
Yes I only have Reaper installed on my Mac - so I only cant test it with it.

OSX is 10.11.4 El Captain.

elieserdejesus:
Hi @vector21 , can you help us testing the new beta? If yes [this is the link](https://github.com/elieserdejesus/JamTaba/releases/tag/v2.0.19-beta)

Now JamTaba Standalone can load your AU plugins, and JamTaba can be used as an Audio Plugin too.

vector21:
Link is not working... :-)

elieserdejesus:
Hi @vector21 , this is the all releases link: [https://github.com/elieserdejesus/JamTaba/releases](https://github.com/elieserdejesus/JamTaba/releases)

vector21:
Chat works as expected now! Thanks for fixing 👍

elieserdejesus:
thanks @vector21 !

PR #691, opened 2017-02-01
[WIP] Modal text editor

elieserdejesus:
Replacing (In AU and VST plugins ) text fields with a invisible top level dialog containing another text field. Using this trick (text field inside a top level window) hosts are not eating keystrokes. 

The old code was using low level keyboard hooks in windows. This solution is possible in Mac too (AU Plugin), but users need activate some security privileges to allow low level keyboard hook (bad thing).  

Many threads in JUCE forum are discussing the host->plugins keystroke problem, and the most reliable, multi plataform and multi DAW solution is a text field inside a top level dialog.

This will solve #668, #661, #657  

- [x] Replace channel name text field
- [x] Replace chat text field
- [x] Replace user name text field
- [x] Check if comboboxes are working fine without keyboard hook.
- [x] Try to solve the issue when the MainWindow is moved and a QLineEdit is focused.
- [x] Avoid leaking in text modifiers
- [x] This solution is not working in mac yet


elieserdejesus:
@jonjamcam , i release a 2.0.19 beta [https://github.com/elieserdejesus/JamTaba/releases/tag/v2.0.19-beta](https://github.com/elieserdejesus/JamTaba/releases/tag/v2.0.19-beta)

Can you use this beta for some days, please? I change a lot of things in VST plugin to solve the key pressing problems in VST and AU plugin. Some hosts are "eating" the keys, plugins never receive these key strokes. I don't know any clean and beatutifull solution, all solutions are very hacky. The solution I'm using now is create a "fake" input text field in top of the "real text field". This "fake" text field" is inside a new Window (a QDialog), but this window is transparent/invisible. For users not is changed, but when we create a new Window hosts can't "eat" the keystrokes. The solution is very hacky and is not perfect. You you notice small changes in text and combobox fields, but it's all aceptable. 

I tested in Reaper (Windows and Mac) and Logic 9, VST and AU plugin working ok. I tested in Sonar too and the result is terrible. I will work in this in another round, Sonar is not a high priority at moment.

jonjamcam:
> Can you use this beta for some days, please?

Sure. I'll use it all along.

Issue #712, opened 2017-02-08
[Looper] Record own or others single Track while jamming in Jamtaba 


MagnusDrumsTrumpet:
Hey i had the idea for a Record Funktion on each Track individuali in Jamtabe while jamming
 to for instance loop it back to build up an jamm
lets say i play drums hit the reck button
 it recs and if i am ok with it i press loop  and it should place the last complete bpi drumtrack recorded
in the jam
 the i could play some chords onthe fly and use them also in same way
 or i could loop chords fron the guitarist to loop them back that ge can improvise with it
 well 
i hope this is understandable 
maybe it is possible to save then on several register to hold them and choose what is needed in the jam
so  what do you think)
 greetings from Mag

elieserdejesus:
@MagnusDrumsTrumpet , I think this is nice idea. Sometimes I miss this feature too.

We need discuss more details to avoid rework when writing the code. @jonjamcam , are you interested in discuss this feature details too?

Starting the discussion:
0 - This idea can be solved using an external VST looper plugin. So, I need ask: work in this feature (is not easy and quick) worth it?

Assuming the answer is yes:

***
1 - I have the idea to JamTaba native plugins. I think this feature can be the first cool native plugin: **JamTaba Looper**.

2 - We can work with a fixed loop length (using current bpi value) or let user choose the loop length. For example **Loop length = 1** means *one interval*, **Loop length = 2** means *2 intervals*, etc. 

3 - Allow record and loop the track input (audio or midi)

4 -  Magnus mentioned *"...or i could loop chords fron the guitarist ..."*. This can be more complicated to implement. A first idea in my mind is: The Looper plugin will contain a **input** combo box. This combo will show the options *audio, midi, no input, and one entry for every ninjamer channel*, so will be possible record and loop **every incomming data**, from local or remote inputs.

5 - Some ideas for cools User interfaces? Some cool "looper" plugins or software?

qulf:
Nice idea but dont forget abaut midi suport for looper and stereo pan for looper also with midi lern option.  I like lopper from guitar rig like becuse that is one witch work with me. :) I try others vst but always was something not good with them.something like guitar rig looper but with four butons start/play/records for four inpedence loops plus stereo pan. That is my idea. Cy on jam. :) sorry for language

qulf:
[...]

PR #721, opened 2017-02-19
[WIP] Looper

elieserdejesus:
- [x] Fix the mix when 'overdub' and 'hear all' are both checked. The local audio is doubled in the mix.
- [x] Magnus reported: if you use the loops and then change the bpm reaper crashes,
- [x] volume fader in each layer
- [x] pan fader in each layer
- [ ] Auto discard not completed loop layers when recording in replace mode.
- [ ]  save the loops and use them later
- [x] "play locked tracks only" checkbox in all layers mode too.
- [ ] allowing mode to be stored as an entry in the json file.
- [x] If you uncheck overdub while recording using overdub checked, Jamtaba crashes in all modes.
- [x] "hear all" during recording checkbox is available in "all layers" mode. Isn't it redundant?
- [x] If the lock button is used in the first layer only in sequence/all layers and you record a second time and the first layer is locked, recording on unlocked layers won't stop. Unlocked layers will be replaced over and over again (I posted this one before, but maybe you didn't see it)
- [x] Only current layer is recorded when overdub is checked in "sequence" and "all layers". Same happens in random mode so is it not possible to choose the layer to be overdubbed. One idea to solve this is **allow layer selection while playing when overdub checkbox is checked**.
- [x] an overdub checkbox in all layers mode

- [x] Record the same layer again: add a record mode feature: replace and overdub. Currently the looper is working in replace mode only.

- [x] a lock icon/button for each one to protected the layers  (when recording). 

- [x] double a Loop from 8 to 16 or 32

- [x] Fix subchannel index order, see https://github.com/elieserdejesus/JamTaba/issues/712#issuecomment-281251387

- [x] Use yellow color for waiting. This can be applied to play and record buttons.

- [x] Use red color for record and green color for play when active.

- [x] More contrast of the wave color

- [x] Change boost values order to 0 -12 +12.

- [x] 2 new play modes: Locked and unlocked layers (locked layers implemented as SEQUENCE mode option)

- [x] Allow layers already recorded to be heard as you record new ones (typical overdubbing)

- [x] When recording a second time during the wait state the layer about to be recorded is muted. The expected behavior is to be able to listen to it's content in all layers/selected modes. In the case of the random/sequential mode the layer you are hearing before pressing record a second time (the highlighted one) is the one being muted. The idea is to be able to listen to it complete during waiting state too.

- [x] In selected mode if you have several layers and you record a second time using overdub (over one of them) then you will hear all layers at the same time. This does not happen if you don't check "overdub" (correct behavior). **1) What is expected is to hear only the selected layer, not all the others**. 2) include the "hear other layers" checkbox in the selected mode too, so there will be two checkboxes: "overdub" and "hear other layers",


jonjamcam:
@elieserdejesus I got some compile errors in mingw32

image

elieserdejesus:
@jonjamcam , It's fixed in last commit.

jonjamcam:
cool now. thx @elieserdejesus .

jonjamcam:
@elieserdejesus Just tested the new play/rec toogle buttons. when play is activated automatically (after recording) is not possible to stop. The toogle works OK only when first pressing the button.

elieserdejesus:
thanks for reporting @jonjamcam , these 2 issues are fixed now.

jonjamcam:
@elieserdejesus I like the new boost button style, but I think a **0 -12 +12** order is better than the actual **0 +12 -12**. It happened to me when I tried to lower a very loud player it first boosted him +12 and that was a very unpleasant experience (to say the least :D).

elieserdejesus:
> I think a 0 -12 +12 order is better than the actual 0 +12 -12.

I agree @jonjamcam , sorry for your bad experience.

qulf:
Another suggestion. I using jamtaba vst with reaper. When I use  lopper I looped all chanels from reaper. Mayby more chanels with vst version? On that moment we have two. But for looper is nice to have one for drum, one guitar, bas, keys. I personally need four :). And another but question. How to route chanels to jamtaba vst in reaper? I know haw to route all chanels to jamtaba but if I wont use two jamtaba chanels one for drum another for guitar I dont know how to do. Help...:)

jonjamcam:
> more chanels with vst version? On that moment we have two.

@qulf The number of channels is limited by the server. Public servers allow maximum two channels per user . Some admins allow more channels, but normally this is done in private servers.

> But for looper is nice to have one for drum, one guitar, bas, keys. I personally need four :)

This will be very easy to do when overdubbing/lock is implemented. 

> I know haw to route all chanels to jamtaba but if I wont use two jamtaba chanels one for drum another for guitar I dont know how to do. Help...:)

In reaper you need to do two things:

1.- open Jamtaba vst and click the **plugin pin connector button** on the upper-right corner

image

You get this window:

image

Now press the **+** button on the left to add more **input channels**:

image

This is done to allow Jamtaba to receive two stereo channels:

image

2.- Go to reaper's mixer and route any reaper channel to Jamtaba input 3/4:

image

Use the **parent channel combo box**:

image

qulf:
Thanks jonjamcam. That is helpfull.

jonjamcam:
@elieserdejesus The lock function works perfect! It's easy to navigate as expected.

Question: what do you think about adding two additional play modes to use it more intuitively

1.- play mode: locked layers (only)
2.- play mode: unlocked layers (only)

First I thought about adding mute and solo buttons for each layer to be able to preview them more accurately (https://github.com/elieserdejesus/JamTaba/issues/712#issuecomment-281685543 number 3), but using the above play modes may have the same effect.

elieserdejesus:
@jonjamcam the play modes **locked and unlocked layers** are good ideas and not difficult to program.

About the recording when **selected play mode** is activated: I agree with yout, the behavior you described is more intuitive than current programmed behavior. But I think is not good **select a play mode** and observe side effects in recording, because we are controlling the recording behavior with a "play mode" control. The semantic is not ok for me. I think the problem is this separation between "play" and "record" modes, maybe this is the wrong way. Maybe the looper  have just "one big" state, and not separated play and record states. This is not clear enough in my mind at moment, I need spend more time thinking about this "looper state" stuff and how this idea could be translated in code.

jonjamcam:
> Maybe the looper have just "one big" state, and not separated play and record states. 

I think replace mode can be ditched if you allow individual layers to be deleted using a red X or something (as easy to use as the lock button). This way the looper will work in overdub mode all the time and the "replace" function is done manually.

jonjamcam:
I have another idea (the opposite of what I wrote in the comment above):

This is a way to do overdubbing much simpler and keeping flexibility using the way the looper is working right now:

If you allow **layers already recorded** to be heard **as you record new ones**, then you have typical overdubbing, but with the possibility to edit each layer (this is not possible to do with the majority of loopers, specially pedals. The audio is hard-mixed in these cases).

So for example if you use _all layers_ mode. First layer is recorded. When second layer starts to be recorded, you can hear the first one. Then when the third is recorded you hear 1 and 2 too. For the next round, recording will occur only in unlocked tracks and data will be replaced (as it is now). You can call this mode "default" instead of "all layers".

 _sequential/random/selected_ modes will work as usual.

elieserdejesus:
> If you allow layers already recorded to be heard as you record new ones, then you have typical overdubbing

@jonjamcam , I will program this idea (it's a quick task) and you can test if the general behavior is more intuitive.

You mentioned "delete layer" in previous comment. At moment DELETE is working to erase the current layer content. When you said "delete layer" are you thinking in **delete layer** or **erase layer content** and keep the empty layer?

elieserdejesus:
@jonjamcam , I added a check box to allow "hear layers while recording":
image

jonjamcam:
> When you said "delete layer" are you thinking in delete layer or erase layer content and keep the empty layer?

The second one: erase layer content and keep the empty layer. A quick way to discard bad takes.

jonjamcam:
I tested the "hear layers while recording" feature. It works as expected during recording. But once recording is finished, during play, I think the logical behavior is to continue hearing all tracks. At the moment the default mode is _sequence_, so in this case when playing you hear layers one by one. 

as I see it:

1.- "hear layers while recording" ON is naturally linked to "all layers mode".
2.- "hear layers while recording" OFF is naturally linked to "sequence/random mode".
3.- I'm not sure about how "selected mode" relates to this funxtion yet.

elieserdejesus:
Nice synthesis @jonjamcam . I'm thinking in these changes:

1 - Change **Play Mode" to just **Mode**. The selected **mode** will apply both to recording and play.

2 - The **hear layers while recording** checkbox will be visible ONLY in **all layers** mode, and checked by default.

3 - A new **overdub** checkbox will be visible ONLY in **selected** mode.

Overdub and *hear while recording* will not be available in **sequence** and **random** modes.

What is your opinion?

jonjamcam:
> What is your opinion?

sounds good to me :D

elieserdejesus:
@jonjamcam , the changes discussed in my previous comment are done. I think now the looper usage is nice, and I hope all these possibilities (modes, check boxes, etc.) are intuitive enough :)

jonjamcam:
@elieserdejesus can't conpile in mingw32:

image

elieserdejesus:
@jonjamcam , mingW compilation is fixed in last commit.

jonjamcam:
Works great @elieserdejesus! I think it's a pretty flexible looper now and easy to use (hope it's easy for others too!) Will you update the beta so others can test it?

I'm thinking it's better to ditch the retrospective recording feature. The looper is very complete as it is now, but the most important reason: I see you did a lot of coding so far and it can save you working hours if we eliminate it. The feature is very low priority.

jonjamcam:
One behavior that is not what you would find in other loopers:

 When recording a second time during the wait state the layer about to be recorded is muted. I think the expected behavior is to be able to listen to it's content in all layers/selected modes. I'm not sure about the Random/sequential modes. They seem to work fine as they are now in this regard.

jonjamcam:
One last thing: In selected mode if you have several layers and you record a second time using overdub (over one of them) the you will hear **all layers** at the same time. This does not happen if you don't check "overdub" (correct behavior).

elieserdejesus:
>Will you update the beta so others can test it?

Yes @jonjamcam , I will update soon, just finishing some not too complicated tasks in this PR.

> I'm thinking it's better to ditch the retrospective recording feature. 

Shure, it's a cool feature, but we can delay this feature to next release.

I will check another 2 issues you mentioned.

jonjamcam:
I found an issue when the **lock button** is used **in the first layer only** when using sequence/random/all layers . When you record a second time and the **first layer is locked**, recording on unlocked layers won't stop. Unlocked layers will be replaced over and over again.

jonjamcam:
from https://github.com/elieserdejesus/JamTaba/pull/721#issuecomment-282556795 :

> In selected mode if you have several layers and you record a second time using overdub (over one of them) the you will hear all layers at the same time (overdub checkbox is working the same as "hear layers while recording"). This does not happen if you don't check "overdub" (correct behavior).

> What is expected is to hear only the selected layer, not all the others.

I gave this issue a second thought. My reasoning for the expected behavior posted here was that recording and play should work in the same way. So if during play you can only listen to the selected layer, then it made sense to me that recording behaved the same way.

But maybe this was an incomplete solution. I think a more flexible solution is to include the "hear other layers" checkbox in the **selected mode** too, so there will be two checkboxes: "overdub" and "hear other layers", I think there are times when you'll need both options.

what do you think @elieserdejesus ?

elieserdejesus:
>  include the "hear other layers" checkbox in the selected mode too

Sounds good to me @jonjamcam , this is not complicated to change.

elieserdejesus:
@jonjamcam , the "hear other layers" checkbox is added in the selected mode too. Very trick to get all modes working fine :)

jonjamcam:
@elieserdejesus I tested last commits and found one issue with the "selected layer" mode:

 When "hear other layers" and "overdub" are both checked a strange behavior occurs after the first recording onwards: the audio from the layer **that was already recorded** is muted. It's **not erased**, only muted (you hear only the incoming audio), but the overdub is succesfull, meaning **the audio is recorded as expected**. You can do a simple test to see what happens:

1.- with all layers empty select "selected layer" mode and check both "hear other layers" and "overdub".
2.- now start recording. You'll notice that from the second recording (first overdub) on, you can't listen to the audio already recorded.
3.- If you stop recording and play, you'll see the overdub was done as expected.

elieserdejesus:
@jonjamcam , I fixed the overdubbing issue you mentioned before.

I dropped the RANDOM mode. Now we have just 3 modes: **SEQUENCE, ALL LAYERS and SELECTED**. 
Now **random** and **locked layers** are options in SEQUENCE mode. So you can play a **random sequence**, or a **locked layers random sequence**, etc.

jonjamcam:
@elieserdejesus There's one little issue stil pending (https://github.com/elieserdejesus/JamTaba/pull/721#issuecomment-282559805)

>I found an issue when the lock button is used in the first layer only when using sequence/random/all layers . When you record a second time and the first layer is locked, recording on unlocked layers won't stop. Unlocked layers will be replaced over and over again.

jonjamcam:
@elieserdejesus The selected layer mode is working very intuitively, but there's something missing when using it: **you can't hear other layers when playing**, for example when you finish overdubbing and you switch to play mode. In this case the only way to listen to all layers is change mode to "all layers", which is inconvenient, specially if you want to select another layer and continue overdubbing.

I think a simple solution for this is change the "hear other layers **during recording**" checkbox function for a "hear other layers **during play and record**" or just "always hear other layers". This will add the missing feature.

> I dropped the RANDOM mode. Now we have just 3 modes: SEQUENCE, ALL LAYERS and SELECTED.
Now random and locked layers are options in SEQUENCE mode. So you can play a random sequence, or a locked layers random sequence, etc.

I agree it's much better now. You can implementing the "play locked tracks only" checkbox in **all layers** mode too. 

Last thing, what do you think about allowing **mode** to be stored as an entry in the json file? I'm thinking users will mostly stick to the same preferred mode  after restart.

elieserdejesus:
>  change the "hear other layers during recording" checkbox to "always hear other layers"

@jonjamcam, I think this can be confusing, becase you are using SELECTED mode but hearing ALL layers. I think you are thinking in SELECTED mode as a recording mode only, right? I think your suggestion make sense as a "recording" issue, but is confusing as a "playing" mode. I'm not shure how to solve at moment. 

> "play locked tracks only" checkbox in all layers mode too.

Perfect!

> what do you think about allowing mode to be stored as an entry in the json file? I'm thinking users will mostly stick to the same preferred mode after restart.

I totally agree, I will add a task for this feature.

jonjamcam:
>  I think this can be confusing, becase you are using SELECTED mode but hearing ALL layers. I think you are thinking in SELECTED mode as a recording mode only, right? I think your suggestion make sense as a "recording" issue, but is confusing as a "playing" mode. I'm not shure how to solve at moment.

I understand your argument. I thought about it too. "selected mode" =/= "hear all layers".  So I thought about an alternative way to do this more consistent with the modes: tell me what you think @elieserdejesus about implementing an overdub checkbox in all layers mode. You get roughly the same result.

The idea is to be able to overdub one selectd track while listening to all others/locked layers,

elieserdejesus:
> an overdub checkbox in all layers mode

This make sense @jonjamcam . I will try this approach.

jonjamcam:
@elieserdejesus Did some testing of the last commits you made. The new design is very nice. I guess you're probably still testing the new changes, but here are some things I found to consider:

1.- If you **uncheck overdub** while recording using overdub checked, Jamtaba crashes in all modes.

2.- "hear all" during recording checkbox is available in "all layers" mode. Isn't it redundant?

3.- If the lock button is used in the first layer only in sequence/all layers and you record a second time and the first layer is locked, recording on unlocked layers won't stop. Unlocked layers will be replaced over and over again (I posted this one before, but maybe you didn't see it)

4.- Only current layer is recorded when overdub is checked in "sequence" and "all layers". Same happens 
in random mode so is it not possible to choose the layer to be overdubbed. One idea to solve this is allow layer selection while playing when overdub checkbox is checked.

Please take the time to watch this video. It's a great example how you manipulate layers in real time and it's kind of standard technique:

https://www.youtube.com/watch?v=izrL7pBdkaw

I think one good goal could be that Jamtaba looper can do this kind of layer muting/overdubbing using the mouse.

elieserdejesus:
> I guess you're probably still testing the new changes

Yes @jonjamcam , I'm fighting with these Modes + options logic puzzle :)

The 4 issues you mentioned are very important, thanks. I'm putting these 4 issues as tasks in this PR.

elieserdejesus:
@jonjamcam , I worked in some tasks in this PR. Can you check this new version, please?

I need your help (if you have some time) testing the looper when BPM and BPI are changed. Now when BPM is changed the looper is stopped (fix a bug mentioned by Magnus).

I see a problem when changing BPI but I can't discover the pattern to reproduce the bug. Sometimes when we change the BPI the layers are not streched/copyed. I was testing the VST plugin and when the interval reached the last beats (empty samples in looper) Reaper auto mute the JamTaba track. The meter in JamTaba show +127 dB. I think when BPI change (enlarge) some trash values are stored in the end of layers, and if these values are mixed the result is a very high audio peak, and Reaper is auto muting.  If you can test this issue please take care of your sound system and your ears.

jonjamcam:
> I see a problem when changing BPI but I can't discover the pattern to reproduce the bug. Sometimes when we change the BPI the layers are not streched/copyed.

@elieserdejesus I did several tests **in standalone version** changing BPI. As far as a patter is concerned, I found the following behavior:

1.- Close Jamtaba and launch again (very important), enter a server and record your layers (but **don't change BPI**). Once finished, test the stretching (now you can change the BPI): if you change the BPI to any number the stretching of the wave works perfectly (no silence added).

2.-  To test the added silence: **don't restart Jamtaba**, delete all your layers one by one using the discard button and record them again using any BPI you want. Now test the stretching changing BPI: the silence at the end appears when changing to a higher BPI every time.

My hypothesis here is: only if you record your layer **after** BPI has changed, then the silence appears. To avoid this Jamtaba should be restarted as only option. If you leave the server and enter again, you'll have the same problem of silence added.

I hope you can reproduce the bug as I did in my system. 

> The meter in JamTaba show +127 dB.

 Using standalone version I didn't get saturation in Jamtaba meter. If you send me the VST plugin I will sure test if it happens in my system too with reaper.

elieserdejesus:
thanks @jonjamcam , I hope it's fixed now.

jonjamcam:
>I hope it's fixed now.

@elieserdejesus I tested the case above and it looks solve to me too.
***
 I did some tweaking of the code that handle layer selection. I changed some things to make it easier to play live (for my taste :D). Another thing I think it's not good to allow layer selection while recording/waiting. What's your your opinion?

image

elieserdejesus:
> not good to allow layer selection while recording/waiting

shure @jonjamcam , I will change the code because is not possible (I suppose) create a new PR based in this PR.

jonjamcam:
@elieserdejesus Did some work on the looper colors to match the corresponding palette:

[All Themes_looper palette match.rar.txt](https://github.com/elieserdejesus/JamTaba/files/815952/All.Themes_looper.palette.match.rar.txt)

Not finished yet, but I'm close :D

elieserdejesus:
@jonjamcam , in last commit I changed the signal path as you suggested.

I will check these CSS issues.

qulf:
Good work. I play little bit and first impression is nice.

elieserdejesus:
@jonjamcam , about the CSS issues, in **BaseTrack.css** change
```
BaseTrackView #panSlider::handle {
     ... rules
}
```

to just

```
#panSlider::handle {
         ... rules
}
```
The first syntax will apply the rules to *all BaseTrack child controls named as "panSlider"*. The second syntax will apply the rules to **all controls** (inside a track or not) named as "panSlider".

***
About the looper button: add this rule in **LocalTrack.css** file:
```
LocalTrackView #buttonLooper:disabled
{
    background-color: red;
}
```

***
I have a suggestion: please try reduce the contrast in Record and Play group options rectangle. At moment these rectangles are a bit "over". These rectangles are capturing the "eyes attention".
image

jonjamcam:
> please try reduce the contrast in Record and Play group options rectangle. At moment these rectangles are a bit "over". These rectangles are capturing the "eyes attention".

@elieserdejesus is this a right amount or you're thinking less?

image

elieserdejesus:
@jonjamcam , I think this rectangles are very good now, thanks for the changes.

jonjamcam:
> thanks for the changes.

@elieserdejesus just incorporate these changes into this PR if you feel it's ready. I think there are still little details to correct, but it's fine in general terms.

elieserdejesus:
@jonjamcam , I merged this PR in master, now you can create a PR with your changes.

I created new issues for the pending changes in this PR.

PR #729, opened 2017-03-04
[Looper] Alowing focused layer

elieserdejesus:


elieserdejesus:
@jonjamcam , I implemented the "focused layer" idea you mentioned. It's nice now, very flexible to record.
Can you check if is working as expected in all modes?

jonjamcam:
Works great! THx @elieserdejesus The only issue I see is that you **can** focus locked layers, but you can't record on them. As focus is intended only for recording, locked layers should be exempted from fucus.

elieserdejesus:
@jonjamcam , I fixed the focus in locked layers.

jonjamcam:
cool, it' perfect now! thx.

PR #736, opened 2017-03-08
All themes preparing text color fix

jonjamcam:
Fixing text color for preparing state of Xmit button all themes


elieserdejesus:
thanks @jonjamcam

PR #738, opened 2017-03-08
All themes looper button non xmit color fix 

jonjamcam:
- [x] Rounded theme looper button fix

- [x] Black theme looper button fix

- [x] ice theme looper button fix

- [x] Flat theme looper button fix

- [x] Volcano theme looper button fix

- [x] Rounded theme looper button icon color fix

- [x] Boost button non-xmit unchecked color fix

### DIFF
--- a/src/resources/css/themes/Black_nm/LocalTrack.css
+++ b/src/resources/css/themes/Black_nm/LocalTrack.css
@@ -200,3 +200,9 @@ LocalTrackView #buttonLooper:disabled
     border: none;
     background-color: qlineargradient(x1:0, y1:0, x2:0, y2:0.95, stop:0 rgb(140, 140, 140), stop:1 rgb(120, 120, 120));
 }
+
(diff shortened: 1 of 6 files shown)


jonjamcam:
@elieserdejesus I had a little issue with the **rounded** theme and the looper button **disabled** in **non-xmit** state. The problem seems to be that the circle inside the icon has the same color of the button so it disappears when dimmed:

image

I noticed that the looper icon has sligtly brighter color than the stereo inverse button for example. What do you think is a better solution:

1.- change the "non-xmit" colors for this particular theme.
2.- match the looper icon to be a little darker

elieserdejesus:
>  I fixed the easy way. Changing the background color for that specific button:

Nice @jonjamcam . Can I merge the PR?

jonjamcam:
> Can I merge the PR?

Sure. It's complete.

elieserdejesus:
thanks!

PR #740, opened 2017-03-10
[WIP] Save and load loops

elieserdejesus:
- [x] Resample when necessary
- [x] Implement the "browse" in load menu button
- [x] Add a looper preferences tab to disable/enable encoded audio when saving loops
- [x] Allow custom loops folder in Looper preferences tab
- [x] Save ogg audio if checked in preferences
- [x] Fix wave drawing while using big audio buffer size.
- [x] Allow load audio files
- [x] Save and load locked layers flag
- [x] crash when saving a loop while playing it
- [x] layer number (in the layer dialog box) is not updated when loading loops to the new layer number.
- [x] limit the minimum layers possible to the ones that are actually being used. Replace the spin box with a combo box and disable not allowed values.


elieserdejesus:
@jonjamcam , this PR (in progress) is allowing saving and loading loop layers content.

At moment the locked layer flag is ignored. If you lock one layer, save the loop and reload the layer will appear unlocked. We have 2 options: Save and load locked layers flag for each layers or **always lock loaded layers**? I'm more inclined to always lock loaded layers, because if users saved the loop it's a good material, and make sense lock all layers by default. What is your opinion?

jonjamcam:
>I'm more inclined to always lock loaded layers, because if users saved the loop it's a good material, and make sense lock all layers by default.

The problem I see with locking all loaded layers is: what if someone saved complete project with both locked and unlocked layers. When loading it would be impossible to know which layers were the good ones. 

Sometimes I leave layers unlocked not necessarily because is bad material, but because I'm not using it temporarily.

jonjamcam:
@elieserdejesus I get a crash when saving a loop while playing it. Saving when the looper is stopped works OK.

elieserdejesus:
> it would be impossible to know which layers were the good ones

Good point!

I will check the crash too.

jonjamcam:
Two more little issues I found:

1.- layer number (in the layer dialog box) is not updated when loading loops to the new layer number.
2.- I think limit the minimum layers possible to the ones that are actually being used. At the moment if you have for example 4 layers recorded it is possible to select 1 layer which is not that good in my opinion.

elieserdejesus:
> I think limit the minimum layers possible to the ones that are actually being used

@jonjamcam , what **being used** means in this context? Non empty layers?

jonjamcam:
> what being used means in this context? Non empty layers?

Exactly. For example if you have 3 non-empty layers, then it should be not possible to allow 2 or 1 layers when changing the layers input box. In this case if you want 1 layer only, then you should delete 2 of them or reset. I hope I'm not missing something here.

elieserdejesus:
got it @jonjamcam , I'm just not shure if this **min layer** logic will be clear for users, because sometimes will be possible reduce the layers count and sometimes this will not be possible. I think the problem is the spin box. When the min layers threshold change the spin box don't change. A greyed value, for example, can indicate a "invalid value", but  this spin box controls is limited.

An idea is replace the spin box with a combo box and disable (grey tone) the invalid values, this will comunicate better the min layer threshold.

jonjamcam:
> I'm just not shure if this min layer logic will be clear for users, because sometimes will be possible reduce the layers count and sometimes this will not be possible. 

True. I wasn't so sure this was a real "problem". It was just a bit uncomfortable for me because some layers appear "hidden", but maybe we need more ninjamer's feedback on this issue.

> I think the problem is the spin box. When the min layers threshold change the spin box don't change.

I kind of like the spinbox. What do you think about discussing this issue with Magnus and Qulf? Good to know their opinion on this one.

elieserdejesus:
> The spin box is not good for the job.

Exactly @jonjamcam , sorry for my Shakespearean english 😄

elieserdejesus:
@jonjamcam, I can't reproduce the issue "**crash when saving a loop while playing it**". Maybe I fixed while coding another task. Can you check if is really solved?

jonjamcam:
> I can't reproduce the issue "crash when saving a loop while playing it"

It's still crashing in my system **only when using my own localhost:2049 server**. No problems when connected to a remote server. Here's how it happens:

1.- Launch Jamtaba
2.- Connect to localhost:2049
3.- start recording until autoplay engages.
4.- press save button
5.- crash.

image

jonjamcam:
@elieserdejesus I did a clean command and recompiled again. At the moment it looks  the crash is gone :D I'll keep testing....just in case.

elieserdejesus:
@jonjamcam , I have some random crashes here. I added some "preventive" code, I hope the crash is fixed now.

jonjamcam:
@elieserdejesus Testing saving and loading some drum loops I played  I find some crackling (clipping/distorted sound) only after loading the loop previously saved. Same BPI, same BPM. The crackling seems to be in peaks (like in the bass drum hits, for example). No  crackling during normal looper play though.

elieserdejesus:
@jonjamcam , I checked your audio file and see the bass drum hits are clipping. So I added a "hard clip" code to avoid this problem. If the incomming audio is too "hot" the peaks will be hard clipped to 0 dB.

jonjamcam:
> If the incomming audio is too "hot" the peaks will be hard clipped to 0 dB.

Does the clipping happen because the looper audio files are 16 bit? I couldn't get the properties shown.

elieserdejesus:
> Does the clipping happen because the looper audio files are 16 bit? 

yes @jonjamcam , the wave files are 16 bit. In 16 bits the max positive value for each sample is 32767. In your audio some samples are bigger than 32767, but in 16 bit world is impossible represent numbers bigger than 32767. As result these big numbers are change to some invalid negative numbers. These values are very far from original sample values, and the audible result is a cracle, because the waveshape is completely messed by these invalid values.

elieserdejesus:
@jonjamcam , I finished the tasks in this PR. Can you test before release a new beta?

jonjamcam:
> Can you test before release a new beta?

I'm trying to make the "loading files" feature to work, but it's not importing some Wav files created with other software. It loads a silent file. But the wav created with the looper are loaded OK (using the load file entry).

An example of a Wave file created by Mobius:

[super loop guitars2-1-1-0.rar.txt](https://github.com/elieserdejesus/JamTaba/files/836825/super.loop.guitars2-1-1-0.rar.txt)

elieserdejesus:
@jonjamcam , your file is a 32 bits wave file. Now JamTaba is loading 32 bits wave files too, tested with your file and is working nice.

jonjamcam:
> your file is a 32 bits wave file. Now JamTaba is loading 32 bits wave files too, tested with your file and is working nice.

Cool. I think all is working as expected for a beta release

PR #752, opened 2017-03-18
[Looper] Looper layers mute button

elieserdejesus:


elieserdejesus:
@jonjamcam , in this PR I implemented the mute button idea (mute/unmute in next interval). Can you test before the merge, please?

jonjamcam:
The mute button woks as expected. Both waiting states are OK.

 The problem I see is how does the mute button relate to the selected layer and sequence modes. It makes sense to me absolutely in all layers mode, but when used in sequence mode the result is one complete BPI interval muted. I don't see any use of the mute in this mode except as a "skip layer" funcition, probably (in sequence mode the mute will skip the muted layer)

In selected layer mode I don't see a point at all of using it.

elieserdejesus:
@jonjamcam , just checking if I understand correctly: the idea is use the mute "wait" state **only in all layers mode**? Other modes will allow "real time" mute, but not "synced" mute?

***
about the "play locked issue": I can see 2 options: 1) disable the mute button when "play locked" is checked; or 2) ignore mute when mixing the samples (the looper content will always be mixed if "play locked" is checked).

jonjamcam:
> the idea is use the mute "wait" state only in all layers mode? Other modes will allow "real time" mute, but not "synced" mute?

 I tested the all three modes:

1.-  what I experienced in sequence mode is that the mute function (both wait and realtime) will always have a "silence effect" for whole BPI. I can't figure out any musical usage of this. But maybe I'm wrong.

2- in selected layer mode I couldn't find any application for the mute button (both wait and realtime) in non selected layers. But again, maybe I'm missing something.

3.- In All layers mode the mute is perfect.

I can't tell right now how to address these issues. I need more testing to have a more solid opinion.

> about the "play locked issue": I can see 2 options: 1) disable the mute button when "play locked" is checked; or 2) ignore mute when mixing the samples (the looper content will always be mixed if "play locked" is checked).

I think the mute button should be operative for locked layers only so you have room to play with them. I see unlocked layers as already being muted when "play locked" is used, that's why I think the mute should be disabled for unlocked layers only.

elieserdejesus:
@jonjamcam , I think the most clean approach is **hide the mute button** in *sequence* and *selected layer* modes. I think hide the mute button is better then just disable.

> mute should be disabled for unlocked layers only.

Got it!

jonjamcam:
> , I think the most clean approach is hide the mute button in sequence and selected layer modes. I think hide the mute button is better then just disable.

Yes. I agree. Much better.

elieserdejesus:
@jonjamcam , can you check this PR again? Now the mute button are visible in **ALL Layers** only and are disabled for unlocked layers if the "play locked" checkbox is checked.

jonjamcam:
> can you check this PR again? Now the mute button are visible in ALL Layers only and are disabled for unlocked layers if the "play locked" checkbox is checked.

Changes working OK @elieserdejesus .

elieserdejesus:
@jonjamcam , I fixed the "mute in background" issue. Can you run a last review before the merge?

jonjamcam:
>  I fixed the "mute in background" issue. Can you run a last review before the merge?

Working perfect now!

elieserdejesus:
thanks @jonjamcam , I will release a new beta.

PR #753, opened 2017-03-19
All themes gain, mute, solo checked non-xmit color match

jonjamcam:
when gain, solo , mute are checked and the channel is in non-xmit mode (local and remote) now you can see a color difference in all themes.


elieserdejesus:
thanks @jonjamcam

PR #763, opened 2017-03-23
[Looper] Showing current loaded loop name

elieserdejesus:


elieserdejesus:
@jonjamcam , In this PR looper window is showing the current loaded loop name. Can you check before the merge, please?

jonjamcam:
> In this PR looper window is showing the current loaded loop name. Can you check before the merge, please?

Looks great!

elieserdejesus:
> So wnat I think is the expected behavior is to hide the name only in the case of layer content and control
What's your opinion @elieserdejesus ?

I agree @jonjamcam , hide the loop name only in a "full reset".

PR #768, opened 2017-03-24
[WIP] Using blinkable button in rec and play looper buttons

elieserdejesus:
- [x] Implement blinkable rec button

### DIFF
--- a/src/Common/gui/LocalTrackView.cpp
+++ b/src/Common/gui/LocalTrackView.cpp
@@ -84,7 +84,7 @@ public:
         }
 
         if (looper) {
-            if (looper->isRecording() || looper->isWaiting()) {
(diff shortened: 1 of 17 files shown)


elieserdejesus:
@jonjamcam , I'm finished the "programable" play button in looper. 

You mentioned use the blinkable button in REC too. Your idea is blink the rec button while recording?

jonjamcam:
> You mentioned use the blinkable button in REC too. Your idea is blink the rec button while recording?

Not while recording, but in **waiting** recording state.

elieserdejesus:
>The Xmit has two colors, so I don't know if it's more complicated to implement

It's a bit more complicated @jonjamcam , I will delay this for the next round.

PR #774, opened 2017-03-27
All themes blinking rec button color match/play button hover fix

jonjamcam:
matching the behavior of the play blinking button and apply it to the rec blinking button. fixed some play button hover missing too.


elieserdejesus:
thanks @jonjamcam !

Issue #782, opened 2017-03-28
[Looper - sequence mode] check/uncheck automatically the "hear all/locked" checkbox when "play locked" is checked/unchecked


jonjamcam:
The problem I see in sequence mode is when you **check the "play locked"** checkbox and you start recording, you don't hear **only the locked** layers and you have to manually check the "hear all/locked"  checkbox too (two step process) which is not intuitive IMO.

Then when you uncheck "play locked", if you keep "hear all/locked" checked you will hear all layers simultaneously, which is not expected in sequence mode. Here you need to uncheck too the "hear all/locked" checkbox manually to get the expected result.

So I think a good solution is to "link" both checkboxes in this mode, so that the "hear all/locked" checkbox gets the same status as the "play locked" checkbox.

PR #785, opened 2017-03-29
[Looper] Linking 'Play locked' and 'Hear All' check boxes

elieserdejesus:


elieserdejesus:
@jonjamcam , this PR is implementing the issue #782 . Can you check if this it the expected behavior in linked check boxes (**play locked** and **Hear all**)?

jonjamcam:
@elieserdejesus I tested the PR and I think there is a problem. If play locked is checked and you start recording (waiting) you can hear all locked layers play simultaneously. This is not expected in ***sequence mode**. they should play individually.

When playing the layers behave as expected.

jonjamcam:
@elieserdejesus The problem described above is not a problem of this impelementation. The same result was happening before this PR . As far as the implementation of THIS PR, the result is OK, but I think the issue is a pending.

jonjamcam:
@elieserdejesus I'm doing some more tests because I have a confusion as a user with sequence mode. All the features are superimposing in my mind and maybe I'm not judging them correctly.

jonjamcam:
Ok I have more clarity now. :D sorry for the confusion.

 The implementation of this PR is OK. The sync checkboxes work as expected @elieserdejesus .

jonjamcam:
@elieserdejesus Can you help me with this? I need a second opinion about the "sync checkboxes" feature. It works perfect in some cases, but sometimes can have very awful results. In your experience is it better to have both checkboxes separate or in sync?

jonjamcam:
@elieserdejesus I'm starting to think this feature is a drawback instead of an enhancement. I tested during the day and I found with the linked checkboxes we have now one feature that is not possible (as it was before) and that is allow play locked layers one at a time.

 I'm really sorry I didn't do thorough testing before posting the issue. My bad. I hope I didn't make you waste valuable time.

elieserdejesus:
> I'm really sorry I didn't do thorough testing before posting the issue. My bad. I hope I didn't make you waste valuable time.

@jonjamcam , no problem. 

Honestly, I don't like these record and play options, I think is confuse because we have 3 modes, and a lot of combinations with these checkboxes. Some checkboxes are available in some modes but not available in others. Users will need learn a lot to understand which options are **really** available in each mode.  I think the current approach is good enough at moment, but we can think in some more simple in future. Maybe "merging" these current modes and options and work with modes only, I'm not shure, just a random idea. Last time I tried think in something different from **modes + options** I got zero ideas :) It's a complex task imagine a different approach without loose all the flexibility.

Can I close the PR and the issue?

jonjamcam:
> I think the current approach is good enough at moment, but we can think in some more simple in future. Maybe "merging" these current modes and options and work with modes only

Yeah. I agree the simpler the better.

Issue #784, opened 2017-03-29
[Looper - sequence mode] Make looper start from the beginning when doubleclicking the play button when stopped


jonjamcam:
At the moment in sequence mode if you stop the looper and start it again, it continues from the layer it was before stop, This is a good thing in some cases, but there's no way to tell the looper to start from the beginning if necessary.

 One way to solve this could be I think: that after the looper is stopped if you doubleclick the play button then it forces to play from start.

what do you think about this idea @elieserdejesus ?

elieserdejesus:
@jonjamcam , I think this "double click" will be a **hidden feature**. Possibly many users never will note the double click feature because all other button are "single click".

I think a more intuitive solution will be **allow layer selection in Sequence mode**. So you can click in the first layer (or another layer) and start the sequence from the selected layer. In the end we will have a total of 2 clicks too (one click to select the first layer and another to start the looper).

what do you think @jonjamcam ?

jonjamcam:
> a more intuitive solution will be allow layer selection in Sequence mode. So you can click in the first layer (or another layer) and start the sequence from the selected layer. In the end we will have a total of 2 clicks too (one click to select the first layer and another to start the looper).

It sounds good, but how to differentiate with the "focused layer" for recording? Maybe doubleclick in a layer?

elieserdejesus:
> how to differentiate with the "focused layer" for recording?

@jonjamcam , I think this differentiation is not necessary. For example, assuming looper is stopped: if you click in the first layer and click in play button the sequence will start in layer 1. If you click in the layer 2 and click in rec button the recording will start in layer 2. Rec and play will always start from selected/focused layer, no need to differentiate. 

I'm missing something?

jonjamcam:
> I'm missing something?

I see this works OK when stopped. It should work fine. I was wrongly thinking of it in during play.

PR #786, opened 2017-03-30
[Looper] Make looper start from the focused layer in sequence mode

elieserdejesus:


elieserdejesus:
@jonjamcam , in this PR I implemented the idea discussed in #784. Now is possible start the sequence in the focused layer. Now is possible change the current playing layer in sequence mode. I like this behavior, is possible change the current layer in any moment and the sequence will continue from focused layer.

what do you think?

jonjamcam:
@elieserdejesus I see the layer selection works while playing (similar to slelcted layer mode). The problem (for me) is now I can't select the recording layer without interrupting the normal looper flow. 

I think we can keep this behavior, but change the "recording focus" to double click so we can access both features differently

elieserdejesus:
> The problem I have with the ability to change the playing layer in the middle (as this PR proposes) is that now you can interrupt (willingly or by mistake) one layer at any point and this can lead to non-musical results.

Perfect @jonjamcam , I'm convinced.

elieserdejesus:
@jonjamcam , I pushed a commit in this PR allowing layer selection only when looper is stopped.

jonjamcam:
Works great. Thx @elieserdejesus .

PR #787, opened 2017-03-31
[WIP] fr_FR fix in Qtlinguist / keep original translation

jonjamcam:
Fixing a few strings that make fr_FR not open in qt linguist


jonjamcam:
@elieserdejesus It looks both translations (google and doublebass) are overlapped  now. I'll commit to keep the original translations from doublebass, not google's

elieserdejesus:
thanks @jonjamcam .

PR #788, opened 2017-03-31
keep-original-fr_FR

jonjamcam:
This is a fix to keep the original doublebass french translation that got overlapped with the google transalte version.


jonjamcam:
@elieserdejesus I did a revision and I hope it keeps the original idea Doublebass had in mind. BTW it was probably a mistake for to do the google translation in the first place, but it's fixed now.

elieserdejesus:
no problem @jonjamcam , thanks for the review!

PR #810, opened 2017-04-07
jp_JP, fr_FR and de_DE missing transaltions update

jonjamcam:
completing unfinished string for jp_JP, de_DE and fr_FR transaltions via google translate. Still missing pl_PL and pt_BR


elieserdejesus:
thanks @jonjamcam

PR #811, opened 2017-04-08
Ice theme combobox fix

jonjamcam:
This PR rounds the comboboxes borders for the ice theme to match the overall theme concept where there were still squared (in **preferences** and **looper window**). Pending is the squared private server combobox which I still can't find a way to round the borders.

EDIT: I found the problem with the private server combobox. It's solved now.


elieserdejesus:
thanks @jonjamcam !

Issue #818, opened 2017-04-11
[Looper] Some keyboard shorcuts issues


jonjamcam:
I.- When focusing a layer the "keyboard focus" should be on the "layer window", but sometimes if the **max layer** or **Mode** coboboxes were selected (focused) before, you can't use the numbered keys to select layers because the "keyboard focus" is in the combobox. (I'm struggling to try to explain this :anguished: )

So for example,

1.- Load a loop with several layers
2.- Change the max layers first to any number different than you have now (the keyboard focus is on the combobox)
2.- select one layer using the mouse (now it's expected that the keyboard focus changes to the layer window so you can use number keys 1,2,3,etc. to select layer), but the focus is still in the max layer combobox.

At the moment to get the keyboard focus change to layers it is necessary to click the checkboxes or the play/rec buttons or the save/load/reset buttons and this is inconvenient.

***

II.- Once you got the layers focused, the numbered keys work OK to layer selection, but when you press a number higher than the max layer available, the focus disappears (I think trying to focus a layer that do not exist). I think the expected behavior is that key shortcuts that are higher than the max layer number should be disabled, so the focus stay in the same layer.

One example:

1.- load any loop with few layers (not 8), let's say 5.
2.- change the layers with the numbered keys.
3.- If you press the number 6 key, the focus disappears.

PR #819, opened 2017-04-12
Fixing looper focus + keystroke issues

elieserdejesus:


elieserdejesus:
@jonjamcam , this PR is fixing the 2 issues you mentioned in #818. Can you teste before the merge, please?

jonjamcam:
@elieserdejesus Yes both issues solved.

 One commentary: The first issue:

> I.- When focusing a layer the "keyboard focus" should be on the "layer window", but sometimes if the max layer or Mode coboboxes were selected (focused) before, you can't use the numbered keys to select layers because the "keyboard focus" is in the combobox.

still happens when you select the looper comboboxes , but don't change it's options, so the comboboxes are kept in focus. I think this is no problem because if this happens, the user is probably not paying attention to the layers, but to the combobboxes.

IMO it's working very good now.

PR #825, opened 2017-04-15
Some minor volcano, black and ice style fixes

jonjamcam:
1.- matching some rounded borders missing for Ice and volcano theme.
2.- volcano theme boost font color match to palette when checked.
3.- volcano, ice themes border fix
4.- looper comboboxes disabled border fix for ice, volcano and black themes


elieserdejesus:
thanks @jonjamcam !

PR #829, opened 2017-04-19
Db markers

elieserdejesus:
- [x] Showing 0 dB markers in audio sliders
- [x] Showing center marker in pan sliders
- [x] Showing db markers in audio meters

- [x] - When using RMS meter only there's no max peak indicator now.

- [x] - Fix meters in local tracks when collapsed

- [x] - Show less dB markers


elieserdejesus:
@jonjamcam , in this PR I'm implementing some visual improvements. Now we have **0 dB markers** in audio sliders, **center markers** in pan slider and **dB markers in audio meters**.

I tested the new meter in comparison with Reaper meter. JamTaba meter is not too accurated like Reaper, but is nice enough :)   Can you test before the merge?

image

jonjamcam:
That was some big change! Some things I found (this is my laptop screen):

1.- The font color of the dB markers in ice theme is a bit too dark:

image

2.- Same happens in other themes with some RMS meters, like Volcano:

image

flat:

image

Black theme looks OK.

3.- When using RMS meter only there's no **max peak** indicator now. Was that on purpose?

4.- Lowering the local fader does not have impact on RMS meter, only on Peak meter in **second subchannel** (first subchannel works OK)

image

elieserdejesus:
@jonjamcam , please try a clean compilation. In your images all dB markers are black, but I changed these colors for each available theme. I think the new CSS files are not exported in your machine. Try a full recompilation or delete the themes folder to see the correct colors.

I created remaining tasks for the other 2 issues you mentioned.

jonjamcam:
@elieserdejesus I did a clean-runQmake-Rebuild. I also deleted the themes folder to be sure, but the markers still appear black in all themes. I noticed that the entry **Peakmeter** is commented out in all themes:

image
*** 
One good news is that the 2nd task:

> Lowering the local fader does not have impact on RMS meter, only on Peak meter in second subchannel (first subchannel works OK)

Is not happening now. The second subchannel works fine after clean.

elieserdejesus:
@jonjamcam , I fixed the issues you mentioned:
- Collapsed tracks are fine now
- Drawing dB marks behind peaks
- After -24 dB the marks are jumping 12 dB to avoid polute the meter:

image

jonjamcam:
> I fixed the issues you mentioned:

Collapsed tracks are fine now
Drawing dB marks behind peaks
After -24 dB the marks are jumping 12 dB to avoid polute the meter:

@elieserdejesus I think the result of drawing dB marks behind the peaks does not look as good as I thought. The level meter is not solid, so you can see some parts of the marks behind. I think now that you reduced the number of marks, the meter is better behind the marks.

What's your opinion?

elieserdejesus:
I agree @jonjamcam . I changed the code to paint dB marks in front:
image

jonjamcam:
The meter looks pretty good now. I couldn't find any other bugs.

jonjamcam:
One little style issue, though: In volcano and Flat, when local channel in not transmitting, the color for the RMS meter is almost the same as the backgroung of the meter:

image

elieserdejesus:
It's fixed now @jonjamcam , thanks for your help.

PR #832, opened 2017-04-21
Fixing file import in locked layers (fix #830)

### LINKED ISSUE #830: [Looper] Importing a file over a focused locked track replaced the locked track
This happens when using the import menu. Tested using the **pre-release** VST plugin version. Couldn't get drag and drop to work using windows explorer in this case so I couldn't test it completely.

EDIT: I suppose that because the locked track focus issue was solved https://github.com/elieserdejesus/JamTaba/commit/cc75e2e73a8b86a2d8becf043234bd5a3bc84bc8 there should be no problem when playing, but still a locked layer it is replaced when the looper is stopped. Tested in standalone with last changes.

elieserdejesus:


elieserdejesus:
@jonjamcam , in this PR I'm fixing the bug when import file in locked layers. Can you check before the merge, please?

jonjamcam:
Working Good @elieserdejesus . Now files can only be imported to non locked tracks via drag and drop and menu.

elieserdejesus:
thanks @jonjamcam .

PR #834, opened 2017-04-23
Fixing locked layers issues

elieserdejesus:


jonjamcam:
Issues II and III still ot solved.
***
Issue II:

Basically the problem here is that you can't hear the input (you hear silence) **when recording** (not when waiting for record) if **play locked** and **hear all/locked** are both checked (modes: **selected layer** and **sequence**).

You can test this very easily:

1.- load a loop with 2 or more layers.
2.- make sure you have a valid input signal and you can hear it.
3.- lock one layer.
4.- focus on any unlocked layer
5.- record.

What will happen is that you'll hear the input signal OK when waiting, but as soon as recording starts, the input is muted.
***
Issue III:

This problem appears only when **layer 1 is locked**. If you focus anyone of layers  3, 4, etc for recording and start recording, it will not stop when the last layer is recorded, but recording will continue from layer2, 3, etc.

elieserdejesus:
@jonjamcam , I changed the tests and the code to "capture" the situations you described. Can you check again please? Sorry if I'm annoying you with these tests, but my brain is melted :) I need any help.

jonjamcam:
@elieserdejesus Issue I and II solved!

elieserdejesus:
@jonjamcam , I added some tests to check the last problem you mentioned (not monitoring through looper layer issue). All tests are passing now. Can you run a last check before the merge, please?

jonjamcam:
Works perfect now @elieserdejesus . Good Job!

elieserdejesus:
nice @jonjamcam , thanks for your help. I will generate a new pre-release in some minutes.

PR #837, opened 2017-04-25
da_DK translation fixes

jonjamcam:
@elieserdejesus I did a revision of the danish translation and corrected some omissions:

1.- matching endings (:, ...,!, etc.)
2.- correct %1 and %2 so they work as expected
3.- fix '&apos;%1&apos' type strings.
4.- other misc corrections.
5.- add missing qtbase_da.qm
6.- update da_DK.qm

NOTE: I had to use Google translate with two strings: **Sync with %1** and **Host sync** because the "extracomment" was translated instead of the corresponding string.


elieserdejesus:
perfect @jonjamcam , thanks!

PR #852, opened 2017-04-29
Updating missing string in danish qtbase.qm

jonjamcam:
the string "Kopiér" was missing in one entry the original qtbase_da.ts

image

Fixed:

image


elieserdejesus:
thanks @jonjamcam !

PR #855, opened 2017-04-29
ice theme looper borders fix to match win10

jonjamcam:
I noticed win10 has thick checkbox borders. This PR should make ice theme looper look better in win10


elieserdejesus:
thanks!

PR #860, opened 2017-04-30
Volcano/Flat themes improving RMS/max peak color

jonjamcam:
Max peak meter in volcano theme when track is not xmitting was invisible, so I used the same color of the RMS meter.

I kept the original color used for the master RMS meter in volcano as it's matching better the overall palette for theme.

Max peak meter in flat theme when track is not xmitting was invisible too, so I changed the color to average the grays


elieserdejesus:
thanks @jonjamcam .

PR #866, opened 2017-05-01
fixing looper width to fit polish translation

jonjamcam:
Polish translation needed a little more space. The max layer combobox was shrunk and did not appear of correct size in some cases:

image


elieserdejesus:
thanks @jonjamcam !

PR #875, opened 2017-05-15
Fix appdmg invocation in mac installer.sh would assume eli user folde…

fab672000:
…r hardcoded path.


elieserdejesus:
thanks @fab672000 !

PR #896, opened 2017-09-10
Feature/metronome accent patterns

pljones:
Hi @elieserdejesus,

I've done almost all of the changes for the Metronome Accent Patterns feature now.

The one I'm stuck on - so I thought I'd ask to maybe save time - is the text input field I'm using in this version for the custom pattern.  For some reason
- I can't type space
- changes do not get notified

I can cut-n-paste from the chat input box into the pattern box, then trigger an pattern update by toggling from Custom to Off and back... but that's a bit annoying ;).

I'm sure it's something obvious I'm missing about the input field - presumably related to it being in a VST.
(Note: longer term, I'm thinking of using the interval progress to allow custom accent entry but that needs some more thinking about - the text input field was intended as a quick way of getting it working for now.)

Comments on the rest more than welcome, too, of course.

Thanks,

-- Peter


elieserdejesus:
@pljones , nice job! Thanks for your collaboration again!

Some issues observed (tested in Standalone only):

- this PR contains many commits related with changes in the clip sort log generator.  If possible please try to create a new PR containing the metronome feature commits only. It's more clean and easy to review in my side.

- The audible accents pattern is not changed when the text is changed in the line edit. I tried ENTER, change the focused widget, no results. I will check this soon, possible is a simple detail in the signal/slot connections.

- I think show the accents lineedit all the time is not a good ideia. The ideal solution (more clean UI) is show the pattern line edit only when the option **custom** is selected in the combo.

elieserdejesus:
> As I was looking through for places where BPI and BPM were used, as I needed BPI, I saw BPM changes "break" the looper. It seemed reasonable to have BPI changes break it too.

The looper is "filling" the content when BPI is changed. Sometimes this is very usefull, for example when BPI change from 8 to 16, looper is just doubling/concating the content.

pljones:
- Yeah, I noticed after I generated the PR that it included stuff from an earlier patch I thought I'd already had committed.  I'll see if I can roll back and re-branch for these changes.

- As I noted in my first comment, the text input field is giving me problems - that's not finished.  I asked if you have any thoughts on why it doesn't work.

elieserdejesus:
> As I noted in my first comment, the text input field is giving me problems - that's not finished. I asked if you have any thoughts on why it doesn't work.

I tested here (In Standalone to avoid VST issues), added some trace code, but the **textEdited** is never called. I see you are wiring 2 QLineEdit signals: textEdited (user changes) and textChanged (programatic changes). I see textChanged is handled outside NinjaPanel and textEdited in handled internally. Maybe one handler is nulling the other? Something like: textEdit is invoked and is calling **setText**, this will fire a textChanged.  Just a guess ...

pljones:
OK, I've actually merged from upstream now (I got it wrong the first time...).

I see the following have been added:
```
+QT       += core gui network widgets concurrent multimedia multimediawidgets
+
+QTPLUGIN += dsengine # necessary to use QCamera inside VST plugin
```
I'm going to have to get that sorted before the build works.

elieserdejesus:
@pljones , maybe I can send you my static compiled Qt multimedia libs, this will save your time. What version of Qt and msvc are you using?

pljones:
I just updated to 5.9.1
```
Qt Creator 4.4.0
Based on Qt 5.9.1 (MSVC 2015, 32 bit)
From revision 60b8712a42
Built on Sep  4 2017 04:10:05
```

elieserdejesus:
I'm using Qt 5.5.1 and MSVC 2013, I suppose the libs will not be compatible.

I compiled just the multimedia module some days ago.  You can download onlye the [qt multimedia module](http://download.qt.io/archive/qt/5.9/5.9.1/submodules/) and compile using your Qt qmake. 

The process is easy, just unpack the downloaded qt multimedia module sources, navigate to the unpacked folder using a command prompt, run the command **[your current Qt folder]/bin/qmake**.  Now compile with **make** and after just **make install**. The compiled static libs will be in your Qt folder.

pljones:
Hm.  It fails to compile with:
```
d:\qt\qtmultimedia-opensource-src-5.9.1\src\multimedia\qtmultimediaglobal.h:43: error: C1083: Cannot open include file: 'QtGui/qtguiglobal.h': No such file or directory
```

Can't see why.

elieserdejesus:
Here I unpacked the qtmultimedia sources along side other modules:
image

Maybe if you try to compile from your Qt root folder? Maybe the include\QtGui\qtguiglobal.h is searched as a relative path from Qt root folder.

pljones:
The include\QtGui\qtguiglobal.h is in the main package.

For now, I'll try going back to 5.5.1, I think, if you've had that working.

elieserdejesus:
> For now, I'll try going back to 5.5.1, I think, if you've had that working.

Good idea @pljones , the static builds are not officialy tested by Qt guys, so the newer versions are always a surprise.

pljones:
OK, now...  I was being really thick and trying to use my non-static Qt to build the plugin, then wondering why the static version couldn't find it...

Next problem: 
```
O:\git\JamTaba\src\Common\video\FFMpegCommon.h:7: error: C1083: Cannot open include file: 'libavformat/avformat.h': No such file or directory
```
That's not part of Qt and not included in Jamtaba (I see no sign of FFmpeg there).  How is the build expected to get hold of it?

elieserdejesus:
@pljones , my bad, I **you need download the pre compiled libs again**, the pre compiled ffmpeg libs and the headers are available: [https://www.dropbox.com/s/qckwsmaqlditwpb/JamTaba-static-libs.zip?dl=0](https://www.dropbox.com/s/qckwsmaqlditwpb/JamTaba-static-libs.zip?dl=0)

pljones:
Ah right.  Let's see :)

pljones:
OK, I'm going to have to give up for now.  I've completely broken my build environment.  Now nothing works.

elieserdejesus:
> OK, I'm going to have to give up for now. I've completely broken my build environment. Now nothing works.

No problem @pljones.

pljones:
I couldn't leave it.

I've got it building again but I'm seeing
```
dsengined.lib(evrcustompresenter.obj):-1: error: LNK2019: unresolved external symbol MFCreateMediaType referenced in function "private: long __cdecl EVRCustomPresenter::createOptimalVideoType(struct IMFMediaType *,struct IMFMediaType * *)" (?createOptimalVideoType@EVRCustomPresenter@@AEAAJPEAUIMFMediaType@@PEAPEAU2@@Z)
```
Does that mean my qtmultimedia wasn't statically built?  Or am I missing something else?

elieserdejesus:
> Does that mean my qtmultimedia wasn't statically built? Or am I missing something else?

@pljones , this error is new for me.  **dsengined.lib** is the lib used by Qt to implement camera in windows. Maybe it's a version problem. I'm using windows 10, MSVC 2013 and Qt 5.5.0.

pljones:
Is it a Windows library?  It may be I do not have it as I only have VS 2013 Express.

elieserdejesus:
[https://msdn.microsoft.com/en-us/library/windows/desktop/ms693861(v=vs.85).aspx](https://msdn.microsoft.com/en-us/library/windows/desktop/ms693861(v=vs.85).aspx)

The function **MFCreateMediaType** in this windows lib **Mfplat.lib**.

Try link this lib in Standalone.pro or VstPlugin.pro:

image

PR #899, opened 2017-09-13
Metronome accent patterns (take 2)

pljones:
Hm.  I've had several more goes cleaning up the git history but I don't seem to be able to explain what I want it to do -- it always ends up like this...

I'll keep trying but I'll also try to leave this branch open now...


elieserdejesus:
@pljones , you know **git cherry pick** command? You can create a new branch from master, and after this      you can **import** a **range of commits** using **cherry-pick**.

PR #903, opened 2017-09-18
update readme.md

jonjamcam:
including detailed instructions on how to build static QT to conpile the vstplugin


jonjamcam:
@elieserdejesus This is only the first raw version including the static QT instructions you posted.

elieserdejesus:
thanks @jonjamcam !

PR #905, opened 2017-09-19
ice theme looper combobox font color fix

jonjamcam:
problem introduced when changing camera combobox font color


elieserdejesus:
thanks @jonjamcam .

PR #922, opened 2017-10-13
black theme remote disabled channels hover fix

jonjamcam:
In Black theme, the hover for disabled channels only was applied to local channels, not remoteones. This PR fixxes the issue.


elieserdejesus:
Thanks @jonjamcam !

PR #949, opened 2017-10-22
:gem: Adding a new 'Navy' theme

elieserdejesus:
image

***
image


elieserdejesus:
@jonjamcam , @Hitrate , I created a new theme "Navy'. Please take a look. I think is good enough to release in the next version. I see some little problems (like the stereo inversion button) but in general it's good enough.  

Some icons (like the stereo inversion button) need some work to appear correctly in all themes.

PR #967, opened 2017-12-05
Fix compilation error with Qt5.6.2 (missing include file) 

fab672000:
Not sure why it also shows my merges of your main branch, but the fix proposal is only the last commit.


elieserdejesus:
Thanks @fab672000 !!!

fab672000:
np, you did a great job recntly too (my merge since few months ago was large!) 
I have more to come  that should fix the midi inputs not recognized on latest jamtab2 build but I need to discuss a bit about some simple design choices we could opt for. 
What's the best place to discuss design here, should I create a 'feature' issue or maybe we chat on fb ?

elieserdejesus:
> What's the best place to discuss design here, should I create a 'feature' issue or maybe we chat on fb ?

Please create an issue, it's better to keep the discussion public and available to others.

Issue #968, opened 2017-12-05
Small design change proposal: greatly simplify RtMidi lib (multiple!) linkage options


fab672000:
Context: 
RtMidi needs special care on each platform to build properly. Besides, each platform introduces more variations like arch and on osx the minimum target sdk that needs to be considered when generating static libs.

Problem: 
It becomes hard to maintain and bugs that are easy to fix just by changing the build options (not changing the lib source code hopefully!) can't be deployed rapidly.

Proposed Solution:
Incorporate *small only* libs source code like RtMidi directly in the app project (we should need to add only one extra .cpp file in the standalone to be able to remove the static lib dependency).
(The JUCE toolkit as an example also solves this problem by generalizing the limiting of the libs dependencies and uses a tool (projucer) that helps incorporates source code directly in your app.)

Thoughts @elieserdejesus and others ?

elieserdejesus:
> Incorporate small only libs source code like RtMidi directly in the app project (we should need to add only one extra .cpp file in the standalone to be able to remove the static lib dependency).

@fab672000 ,  I totally agree, it's a good solution. 

Inside **src** you will see a **Libs** dir. This is the **3rd part** (non static) libs folder.

fab672000:
Thanks @elieserdejesus, I'll propose shortly another pull request applying this change shortly and that will hopefully solve the rtmidi build problems on all platforms, after work hours of course :)

PR #969, opened 2017-12-06
Fix MidiIn in OSX for standalone (and any future use of RtMidi in other projects)

fab672000:
As discussed in issue #968, removed rtmidi static lib dependency.
Took that opportunity to use the latest rtmidi software, now easy to update (only 2 files).
Tested working great on OSX, but should work fine on all platforms.


elieserdejesus:
thanks @fab672000 , working fine!

fab672000:
NP, thanks for the testing feedback @elieserdejesus!

PR #1025, opened 2018-03-07
Lupdate for all transaltions

jonjamcam:
update needed for text on https://github.com/elieserdejesus/JamTaba/commit/4dc1ac422a4cb19683ae725a9da4edd6185faa50 to be translated.

updated qm files too (except for pending transaltions fr_FR and da_DK).


elieserdejesus:
thanks @jonjamcam !

PR #1035, opened 2018-03-12
Showing camera icon/video when local channels are collapsed

jonjamcam:
Proposing this change because when you use the local channels collapsed, there's no way to turn the camera on/off (nor know if video is being transmitted) unless you expand them.

image

### DIFF
--- a/src/Common/gui/MainWindow.cpp
+++ b/src/Common/gui/MainWindow.cpp
@@ -641,12 +641,13 @@ void MainWindow::showPeakMetersOnlyInLocalControls(bool showPeakMetersOnly)
 
     if (cameraView) {
         cameraCombo->setVisible(cameraView->isVisible() && cameraCombo->count() > 1);
-    }
-    if (!showPeakMetersOnly) {
+        
+        if (!showPeakMetersOnly) {
             cameraView->setMaximumHeight(90);
-    }
-    else {
+        }
+        else {
         cameraView->setMaximumHeight(32);
+        }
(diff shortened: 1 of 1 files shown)


elieserdejesus:
nice @jonjamcam !

I have one issue to discuss: **how the small cam preview will behave when using just one channel?** 

In your picture you have **2 channels**, and the **width is enough to show the cam** preview. But I'm worried using just one channel the width is not enough to show the cam preview.

jonjamcam:
> I'm worried using just one channel the width is not enough to show the cam preview.

No worries @elieserdejesus . The size is very small, but it serves the purpose to have feedback from the camera

image

elieserdejesus:
perfect @jonjamcam !

One last issue:

in the line **642** you will see
```
if (cameraView) {
   ...
}
```

This code is checking if the **cameraView widget** was created in jamtaba initialization. **You must move your code inside this block**.  

In your code you are using **cameraView->setMaximumHeight(...)**. If the cameraView is not created (because user has no camera) jamtaba will crash, the code will try access an invalid (not created) cameraView widget.

jonjamcam:
> You must move your code inside this block.

It's done in https://github.com/elieserdejesus/JamTaba/pull/1035/commits/a930263821b21a7d3f6ad64868e5970e6cc899e3 @elieserdejesus and tested.

elieserdejesus:
thanks @jonjamcam , merging the PR!

Issue #863, opened 2017-05-01
Levels positions


Hitrate:
![b41941fe-2dd1-11e7-9c4e-b73d1004994c](https://cloud.githubusercontent.com/assets/26333462/25569817/d30f6cc2-2e1d-11e7-984b-4f16deeebfbb.png)

elieserdejesus:
Hi @Hitrate . I'm not shure if I uderstand the idea in your picture. The current local controls layout is
![image](https://user-images.githubusercontent.com/1012741/31318543-cb53b5e8-ac2a-11e7-97b9-2946a2ef1131.png)

Just confirming: in your picture are you proposing a new local controls layout?

Hitrate:
Yes. The cam icons are way too small imo...Ideally make it auto fit, cause I spend way too much time adjusting the size of jamtaba height and width..it doesn’t remember the previous settings at all, and thus I have to size jamtaba every time it crashes or I remove and reinsert on the fx channnel..so if the local channels are redone so that it takes care of the width problem, ie. currently it’s set for 4:3 1989-2007 webcams, and the cams and channels could be wider when less people are in the channel...and perhaps shrink individual depending on who’s cammed up, thus meed the extra width and who’s not.

I lots of other things to say but let’s leave that for specific topics.

Beat Regards 

Torben Scharling

elieserdejesus:
Slider and dB meters improved in #1044

PR #1044, opened 2018-03-14
New slider meter

elieserdejesus:
This PR is related with #863

- Now the audio meters and sliders are **one widget**. This is more clean (less widgets in the UI). The idea was "stolen" from GarageBand: 
image

- JamTaba new audio meters/sliders:
image

- The old version was using 2 small speaker icons in each audio slider. All these speakers are removed too (clean UI) since the dB marks in sliders are enough to indentify **high** and **low** levels.

***
**Pending tasks**
 - [x] Fix the dB peak colors in Ice theme
 - [x] Fix sliders and meters in Looper window


jonjamcam:
looks good! Tested here succesfully. Found only one detail: game theme slider handle a little wide::

image

elieserdejesus:
> looks good! Tested here succesfully. Found only one detail: game theme slider handle a little wide:

thanks for the feedback @jonjamcam , it's fixed in the last commit.

PR #1057, opened 2018-03-20
troll face completely transparent

jonjamcam:
Troll face was not transparent on the inside. This PR fixes that:

image


elieserdejesus:
Thanks!

PR #1079, opened 2018-04-01
Merge branch 'develop'

fab672000:
- Ensure that standalone window state gets saved restored when maximized as well
- Fix Main Window standalone restore position and size would fail with multiple monitors on OSX and also introduce save restore state high level QT API use.
  Now saving the state allows to persist the toolbars layout as well (i.e. tab bar).
  This new code also does not make low level assumptions on the display and only affects the standalone window to facilitate regression testing.
  Some of this code could be reused in future for the plugin common gui code as well.
- Don't attempt to close a PluginFinder process that has crashed and already exited to avoid hanging and preventing main app to exit.
- Minor line formatting fix in MainControllerStandalone

### DIFF
--- a/src/Standalone/MainControllerStandalone.cpp
+++ b/src/Standalone/MainControllerStandalone.cpp
@@ -446,7 +446,8 @@ void MainControllerStandalone::start()
 
     auPluginFinder.reset(new audio::AudioUnitPluginFinder());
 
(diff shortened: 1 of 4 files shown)


fab672000:
my Qsettings code does not run on windows ... will resubmit

PR #1080, opened 2018-04-01
Fix standalone main window position, size and add state persistency. 

fab672000:
Tested successfully on both Windows and Mac OS X.
Also features minor source reformating and one hang fix in case PluginFinder crashes.


elieserdejesus:
thanks @fab672000 !

PR #1088, opened 2018-04-12
Do not introduce organization name root based settings to keep compat…

fab672000:
- Do not introduce organization name root based settings to keep compatibility with older settings and code. 
- Fix one comment typo.


elieserdejesus:
thanks @fab672000 !!

PR #1092, opened 2018-04-16
Allowing instrument icon in local channels

elieserdejesus:
- Dropping the input text field for local channel names
- Adding instrument icon in local channels
- Changing the channel name "internally" when a new icon is selected (the icon change will be visible to others)


jonjamcam:
> The result is very interesting, works fine with ReaNinjam too (the channel pre defined names are used).

image

yes and you can change the icon during a jam and it updates in the remote as usual. Looking good!

elieserdejesus:
> But I think the remote ICON in jamtaba will not update.

Yes @jonjamcam , changes in remote icons are local only.

PR #1093, opened 2018-04-21
:bug: Fixing dropped intervals (fix #1023)

### LINKED ISSUE #1023: Dropped intervals
All 64bit:
Windows10Pro 1709
Reaper v5.76
JamTaba 2.1.4 2018-02-14

I'm not sure how reproducible this is, so I'll provide a lot of detail, some of which may be irrelevant.

I'd been playing on the public servers for well over an hour, switching servers and using the "Listen" button.

First problem is that the "Listen" button starts failing - the server greys but the circle doesn't appear, neither does the signal graphic and there's no audio.

I'm not sure whether that's related but eventually someone told me I was not transmitting - ninbot "nb levels" agreed... most of the time.  Sometimes, though, I'd get audio through - without disconnecting or anything.

It's hard to detect as it relies on a remote monitor (like "nb levels") spotting it - and, of course, if you stop playing to type that in, you might hit enter after an interval when you really weren't playing.

Unfortunately, this is a major issue.  The purpose of the plugin is sending audio.  Everything else is just added bling factor.  If it doesn't get sending audio right, it's pointless.  Of course, it might not be JamTaba getting it wrong - could be anywhere...

So I'll try to see if I can pin this down a bit.  Any help would, of course, be appreciated.

elieserdejesus:
Fixing https://github.com/elieserdejesus/JamTaba/issues/1023


jonjamcam:
@elieserdejesus Yes Channel 1 is now uninterrupted by video, but channel 2 is :cry: . What do you think if when using video, channel 2 is hidden or disabled?

elieserdejesus:
> What do you think if when using video, channel 2's is hidden or disabled so the user can't use it for audio?

Good idea @jonjamcam .

elieserdejesus:
@jonjamcam , I create a special channel for video. It's an empty channel, I think now is clear enough the channel is used to video only.

Can you run a last test, please?

jonjamcam:
Working ok @elieserdejesus

elieserdejesus:
thanks @jonjamcam !

PR #1103, opened 2018-04-29
Replacing 'trollface' string for 'troll' and adding 'mic' string to autoguessing

jonjamcam:


elieserdejesus:
thanks @jonjamcam !

PR #1115, opened 2018-05-16
Reduce unused variables in asserts

pljones:
Got irritated with these... :)

Unless I'm missing something, there's no benefit in capturing the value to be asserted into a variable before the assert - and, when the assert macro is evaluating to nothing, it leaves the call in place doing something that's not used.  There _should_ be no side effects...


pljones:
I do not understand how this thing works...  Why are all those 2017 commits showing here?

elieserdejesus:
> I do not understand how this thing works... Why are all those 2017 commits showing here?

I don't know too, and it's very complicated to me (no free time) discover the relevant commits in this PR.

pljones:
The "Files changed" link shows the full diff:
https://github.com/elieserdejesus/JamTaba/pull/1115/files

PR #1117, opened 2018-05-22
Reduce "unused variable" clutter

pljones:
OK, second go.

### DIFF
--- a/src/Common/ninjam/client/ClientMessages.cpp
+++ b/src/Common/ninjam/client/ClientMessages.cpp
@@ -387,9 +387,7 @@ ClientToServerChatMessage ClientToServerChatMessage::from(QIODevice *device, qui
 
     QByteArray byteArray(payload, Qt::Uninitialized);
 
-    int readed = stream.readRawData(byteArray.data(), payload);
-
-    Q_ASSERT(readed == payload);
+    Q_ASSERT(stream.readRawData(byteArray.data(), payload) == payload);
 
     auto arrays = byteArray.split('\0');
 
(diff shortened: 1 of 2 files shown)


src/Common/ninjam/client/ClientMessages.cpp
```diff
@@ -387,9 +387,7 @@ ClientToServerChatMessage ClientToServerChatMessage::from(QIODevice *device, qui
 
     QByteArray byteArray(payload, Qt::Uninitialized);
 
-    int readed = stream.readRawData(byteArray.data(), payload);
-
-    Q_ASSERT(readed == payload);
+    Q_ASSERT(stream.readRawData(byteArray.data(), payload) == payload);
```
elieserdejesus: @pljones , can you check if these changes will work when compiling for RELEASE?

From Qt docs:

```Q_ASSERT_X is useful for testing pre- and post-conditions during development. It does nothing if QT_NO_DEBUG was defined during compilation.
```

pljones: Define "work".  The result of the evaluation is assigned to an usused variable.  The Q_ASSERT(xxx) argument simply doesn't appear in the source for RELEASE.

If _side effects_ are required, then they should be commented.  I don't propose to check that.

elieserdejesus:
> Define "work"

**This PR it's not working:**

1 -  I pulled this PR
2 - Compiled in **release** mode
3 - I tried to connect in public servers
4 - *Every server connection failed*

**conclusion**: When compiled for **release** this PR is not working (**impossible connect in servers**). 

From Qt docs:
```
Q_ASSERT does nothing if QT_NO_DEBUG was defined during compilation.
```

The entire code inside Q_ASSERT is replaced by "nothing" when compiling in RELEASE. In this PR some very important socket code are inside Q_ASSERT, and this means the socket is never readed, and the result is **no network communication**.

pljones:
In that case, the original code was very poorly written.

There should have been two pieces of code -- one when QT_NO_DEBUG was defined that simply called the reader method and a separate piece of code that did the assert call.  In either case, the unused variable was redundant.

So please fix the code.  I'll raise a defect.

PR #1118, opened 2018-05-22
Alternate jam directory name format

pljones:
And another go at this one.


DoublebassOO:
I agree the new name format really not convenient ; the classification is not done by date  correctly. Looking for the jam of one precise date is not easy to find.

elieserdejesus:
very nice feature, thanks @pljones!

PR #1147, opened 2018-11-15
use discord bot instead of servebeer

jonjamcam:
I'm guessing discord server has a bot now. if that's the case, this pr solves the problem


elieserdejesus:
thanks @jonjamcam !

Issue #1144, opened 2018-11-12
Standalone: Audio Device revamping needed, OSX not having a device selection. device selection save/restore buggy


fab672000:
Hi guys, I have exciting changes that fix all of this now we have on OSX the device change supported but much more ...

For all platforms: we now get separate audio device input output selection!
and now no default channel for OSX portaudio wrapper code meaning that you can use any input channel on OSX too!

Super excited  toshare all this soon,

@Eli: I also have a patched portaudio lib on OSX that fixes the channel names tyext (that always loose last char), so that now if fully works as expected!

elieserdejesus:
Hi @fab672000, thanks for your collaboration, I have 2 comments:

1 - Please create a Pull Request.

2 - *Now All plaforms get input and output separate device selection*. Maybe I'm wrong, but I think this is very problematic for windows users. In windows you have to choose ONE audio device (for input and output). We have no flexibility in windows like in OSX or Linux.  So, we will force windows users to work like in Mac or Linux, but the windows users database is bigger than others. 

![image](https://user-images.githubusercontent.com/1012741/48361768-5bcefd00-e689-11e8-82a6-ee58de141d7f.png)

[JamTaba Release stats](https://www.somsubhra.com/github-release-stats/?username=elieserdejesus&repository=JamTaba)

fab672000:
Thanks Eli, I will create a PR  as soon as I finish testing and fix a few more things.

Something is telling me we will have more Mac OS X users once they get a device selection back (was disable in OSX and also channel was not selectable) :)

I may be wrong, but I don't think that would be a problem just letting users to have the choice to test different drivers if they would like to but it is still possible (and easy to change) to force the same driver for both input / output in Windows only, if we absolutely need it.
I'll test this with different ASIO drivers on PC and check this as well before I propose a PR, as you would expect :)

elieserdejesus:
> I'll test this with different ASIO drivers on PC and check this as well before I propose a PR, as you would expect :)

Thanks @fab672000 !  Maybe we can have the *output device combobox* disabled (or hidden) in Windows only. So we can force the same I/O device programatically in Windows. 

I'm ok with some code like:

```
#ifdef OS_WIN
      outputComboBox.setVisible(false);
#endif 
```

fab672000:
[...]

PR #1149, opened 2018-11-19
Enhance audio device setup code, enhance settings startup logs, see Issue #1144

fab672000:
As discussed keep single audio device behavior on windows, (still fixes min max channel selection bug in windows too).
UseSingleAudioIODevice constant bool now allows to change the behavior of any platfor, for now Windows default to true and other platforms to false.
Improve settings startup logs.
Fixes issue #1144


elieserdejesus:
@fab672000 , I see you are adding some commits in this PR, please let me know when you are finished here, ok?

fab672000:
Ok, done there is now also the post-cpu profiling first set of optimizations for cheaper peak computations and optimized SampleBuffer underlying code!

elieserdejesus:
@fab672000 , I tested your PR in Windows and is working fine, thanks!

I see just one little issue: *You added the label Input in the combo (see the image), but this label is not the best possibility in Windows (Input and Output in the same combo). The previous JamTaba version was using no label.* 
image

Can you fix, please? Maybe some #ifdef OS_WIN code is enough 😄

fab672000:
added one more commit to fix this, thanks!

fab672000:
My pleasure,  I'd experiment your backup ip locator idea this WE, see if i can help ;-)

PR #1151, opened 2018-11-25
Backup geolocation server implementation,  see #1044

fab672000:
Now allows for 10000 additional request when first server is full, also fixes empty ip's could be santized with .128 postfix.


fab672000:
Ok this PR should diff only 3 files changes, please review this one.

Issue #1146, opened 2018-11-15
Some users do not appear in map


jonjamcam:
![image](https://user-images.githubusercontent.com/15310433/48556548-71336980-e8cb-11e8-8fe3-ce834296b219.png)

elieserdejesus:
@jonjamcam , this can be related with refresh time in JamTaba.

 Maybe the ninjamer *Piw* enter in the server but JamTaba is not updated yet?  You waited some seconds ?   

ps: I'm just trying to eliminate the simple hyphothesys, maybe this is not a bug.

jonjamcam:
I have an hypothesis @elieserdejesus . It looks to me there's a problem related to the cache. Jamtaba is not updating cache for new users, so only "known" users are shown. Maybe there's a problem with ipapi?

elieserdejesus:
I checked **ipapi** dashboard and the free monthly quota is expired:

![image](https://user-images.githubusercontent.com/1012741/48567039-5e2c9380-e8e3-11e8-8160-452086514cb2.png)


![image](https://user-images.githubusercontent.com/1012741/48567138-82887000-e8e3-11e8-964f-0451c187b850.png)


**Too mutch requests, maybe the cache is not working as expected:**
![image](https://user-images.githubusercontent.com/1012741/48567267-c7aca200-e8e3-11e8-803e-fe0800a791dc.png)

elieserdejesus:
Idea: Use more than one **IP to Location** APIs. When one API free plan is exausted (returning some error code) the other API is used. 

[https://ahmadawais.com/best-api-geolocating-an-ip-address/](https://ahmadawais.com/best-api-geolocating-an-ip-address/)

elieserdejesus:
**Just to documenting:** The json returned by **ipapi** when API limits is exausted:
```
{
  "success": false,
  "error": {
    "code": 104,
    "type": "usage_limit_reached",
    "info": "Your monthly usage limit has been reached. Please upgrade your Subscription Plan."
  }
}
```

jonjamcam:
So Jamtaba is using 1000 to 1500 requests a day and probably this number will continue to grow in the future.

Is there a way for each client to obtain this geolocation information from a centralized database, so the requests to ipapi.com are **one per user connected**, not one per client as it is now?

elieserdejesus:
> So Jamtaba is using 1000 to 1500 requests a day on weekends and probably this number will continue to grow in the future.

[...]

PR #1153, opened 2018-11-26
Alternate geolocation server implementation see #1146

fab672000:
Now enjoy a fallback server with 10000 additional requests per month!
Also fixes empty ip's would create false positive errors during geolocation calculation because sanitizeIp would concatenate ".128" to an empty string now creating an non empty string that will not be skipped.

### DIFF
--- a/src/Common/geo/WebIpToLocationResolver.cpp
+++ b/src/Common/geo/WebIpToLocationResolver.cpp
@@ -16,6 +16,8 @@
 #include <QPointF>
 #include "log/Logging.h"
 #include "persistence/CacheHeader.h"
+#include <QSettings>
+#include <QApplication>
 
 using geo::WebIpToLocationResolver;
 
@@ -193,11 +195,20 @@ void WebIpToLocationResolver::requestDataFromWebService(const QString &ip, int r
 
     QNetworkRequest request;
 
+    static QString ipApiKey;
+    static QString ipStackKey;
+
+    if (ipApiKey.isEmpty()) {
+        QSettings settings(QCoreApplication::applicationName());
+        ipApiKey = settings.value("ipApiKey", "22f433857895c84347e949db234af11f").toString();
+        ipStackKey = settings.value("ipStackKey", "7dfb5855025fa60f5ce082a4c376091a").toString();
+    }
+
     QString serviceUrl = "http://api.ipapi.com";
-    QString accessKey("22f433857895c84347e949db234af11f");
+    QString accessKey(ipApiKey);
 
     QString alternativeServiceUrl1 = "http://api.ipstack.com/";
-    QString alternativeAccessKey1("7dfb5855025fa60f5ce082a4c376091a");
+    QString alternativeAccessKey1(ipStackKey);
 
 
     QString lang = sanitizeLanguageCode(currentLanguage);
--- a/src/Common/geo/WebIpToLocationResolver.cpp
+++ b/src/Common/geo/WebIpToLocationResolver.cpp
@@ -141,12 +141,12 @@ void WebIpToLocationResolver::replyFinished(QNetworkReply *reply)
         QJsonObject root = doc.object();
         if (!root.contains("country_name")) {
             if (root.contains("error")) {
-                auto error= root["error"].toObject();
+                auto error = root["error"].toObject();
                 auto code = error.contains("code") ? error["code"].toInt() : -1;
-                if (retryCount<MaxServersAlternatives)
+                if (retryCount < MaxServersAlternatives)
                 {
                     qDebug() << "Error " << code << " received, trying alternative server ...";
-                    requestDataFromWebService(ip, retryCount+1);
+                    requestDataFromWebService(ip, retryCount + 1);
                 }
                 else {
                     qCritical() << "All servers failed, no more alternatives available";
--- a/src/Common/geo/WebIpToLocationResolver.cpp
+++ b/src/Common/geo/WebIpToLocationResolver.cpp
@@ -2,6 +2,7 @@
 
 #include <QJsonDocument>
 #include <QJsonObject>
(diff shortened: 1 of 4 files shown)


fab672000:
I had to create a patch and redo this PR from a hard reset on my master branch, but this time you should only see one commit !

fab672000:
Added few more commits to implement as well the advanced user settings overrides so that developers don't consume all the requests quotas when testing hard ;-)

src/Common/geo/WebIpToLocationResolver.cpp
```diff
@@ -133,8 +138,23 @@ void WebIpToLocationResolver::replyFinished(QNetworkReply *reply)
 
         QJsonObject root = doc.object();
         if (!root.contains("country_name")) {
-            qCritical() << "The root json object not contains 'country_name'";
-            return;
+            if (root.contains("error")) {
+                auto error= root["error"].toObject();
```
elieserdejesus: add a space after "error" (Qt code style)

src/Common/geo/WebIpToLocationResolver.cpp
```diff
@@ -133,8 +138,23 @@ void WebIpToLocationResolver::replyFinished(QNetworkReply *reply)
...
+                auto code = error.contains("code") ? error["code"].toInt() : -1;
+                if (retryCount<MaxServersAlternatives)
+                {
+                    qDebug() << "Error " << code << " received, trying alternative server ...";
+                    requestDataFromWebService(ip, retryCount+1);
```
elieserdejesus: I'm not 100% shure, but I think in Qt code style **retryCount + 1** will be **retryCount + 1**, the plus operator is surrounded with spaces.

src/Common/ninjam/client/User.cpp
```diff
@@ -40,16 +40,13 @@ QString ninjam::client::maskIpInUserFullName(const QString &userFullName)
...
-    QString maskedIP(ip);
-
     // mask ip if necessary to be compatible with ninjam masked IPs
-    if (!maskedIP.endsWith(ninjam::client::IP_MASK)) { // not masked?
-        auto index = maskedIP.lastIndexOf(".");
-        if (index)
-            maskedIP = maskedIP.left(index) + ninjam::client::IP_MASK; // masking
+    if (!ip.isEmpty() && !ip.endsWith(ninjam::client::IP_MASK)) { // not masked?
+        auto index = ip.lastIndexOf(".");
+        if (index) return ip.left(index) + ninjam::client::IP_MASK; // masking
```
elieserdejesus: Separate the conditional in 2 lines is preferred.

fab672000:
I like Upper case for const definitions (they reall are like strongly typed macro when you think about them) As for conditionals, I am following the concept the best code is often the shortest (not syntaxically of course), too many imbricated code constructions are actually considered bad  (check what resharper would suggest as an example for refactoring after code analysis).
We should all use a tool for formatting, on.net apps, I use Code Maid not sure it would work for C++.
This way we don't need to discuss it ;-)-Fab

fab672000:
Talking about code style and what I think is dangerous: I saw many if constructions with no brackets and a carriage return (and did the same to keep the tyle identical) but I think that:
if (a) b; // is OK no risk of bug because of misplaced brackets

```
if (a) {
   b; // OK as if I add there it's within brackets
}
if (a)
    b; // Not OK for me as one could add logic bugs very easily as an example:

if (a)
    b;
    c;
// user intended to add inside an if statement forgetting that the code is actually:
if (a)  {
    b
}
c;

// instead of :
if (a)  {
    b;
c;
}
```

fab672000:
In [http://doc.qt.io/qtcreator/creator-beautifier.html](http://doc.qt.io/qtcreator/creator-beautifier.html)
the also advise for compressed bracketing which i think is not followed in the code base:
```
if (a) {
   b;
} else {
  c;
}
```
I don;t like very much this style, at work we don't 'save' lines and prefer to write:
```
if (a) 
{
    b;
} 
else 
{
    c;
}
```

src/Common/geo/WebIpToLocationResolver.cpp
fab672000: agreed for this one, do you use a code style tool?

src/Common/geo/WebIpToLocationResolver.cpp
fab672000: Agreed

src/Common/geo/WebIpToLocationResolver.cpp
fab672000: Agreed

elieserdejesus:
Thanks @fab672000 . Some time ago I was using a QtCreator plugin to format the code in Qt code Style. The plugin is called Uncrustify, and the config file is here: https://github.com/elieserdejesus/JamTaba/blob/master/PROJECTS/uncrustify.cfg. But if you are using XCode this is useless for you.

src/Common/geo/WebIpToLocationResolver.cpp
```diff
@@ -26,6 +27,9 @@ const quint32 WebIpToLocationResolver::COUNTRY_NAMES_CACHE_REVISION = 1;
 const quint32 WebIpToLocationResolver::COUNTRY_CODES_CACHE_REVISION = 1;
 const quint32 WebIpToLocationResolver::LAT_LONG_CACHE_REVISION = 1;
 
+// Alternative servers private implementation strategies
+const int MaxServersAlternatives = 2;
```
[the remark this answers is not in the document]
fab672000: Could we agree on constants ? IMO it really helps when reading to differentiate them from other variables written with camel convention. I also use member conventions usually (i like the _ prefix to differentiate from stack variables as an example)

fab672000:
Thanks! will make i use this plugin too!

PR #1154, opened 2018-11-27
Applied uncrustified.cfg style to recent contribs I did

fab672000:
Make it pure Qt style!
Also allowed to detected a default without break in Configurator.cpp which is not a bug but would cause a warning.

### DIFF
--- a/src/Common/Configurator.cpp
+++ b/src/Common/Configurator.cpp
@@ -65,12 +65,25 @@ void Configurator::logHandler(QtMsgType type, const QMessageLogContext &context,
 
     QTextStream stream(&stringMsg);
     switch (type) {
-    case QtDebugMsg:    messageType = "DEBUG   ";   messageColor = getDebugColor(context);   break;
-    case QtWarningMsg:  messageType = "WARNING ";   messageColor = COLOR_WARN;   break;
-    case QtCriticalMsg: messageType = "CRITICAL";   messageColor = COLOR_CRITICAL;   break;
-    case QtFatalMsg:    messageType = "FATAL   ";   messageColor = COLOR_FATAL;   break;
+    case QtDebugMsg:
+        messageType = "DEBUG   ";
+        messageColor = getDebugColor(context);
+        break;
+    case QtWarningMsg:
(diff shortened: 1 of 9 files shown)


elieserdejesus:
Hi @fab672000 , I see Uncrustify is breaking some "long" lines in two or 3 lines. In Uncrustify config file the [columns per line is setted as 100](https://github.com/elieserdejesus/JamTaba/blob/cb25e05b40970419d6ba49c1f3a077fe9a8a8a1f/PROJECTS/uncrustify.cfg#L1280). 

What you think about increase this number? I vote to 160 (some people prefer 120).  If you like the idea please change the Uncrustify config file and re-apply Uncrustify in this PR.

fab672000:
I agree and would also prefer more modern 160 columns strategy, will do and update.
Also, what about the Qt style 
`} else {
`
i saw that in code you guys use more

```
} 
else {

```
(I personally even prefer a dedicated line for brackets)

Are you happy with the default (qt style) or should we change this too?

elieserdejesus:
> Are you happy with the default (qt style) or should we change this too?

Please change this too, let's use a dedicated line for brackets as you suggested.

fab672000:
Applied our discussion to uncrustify.cfg, let me know if there is anything else you would like to see on that cfg file before we rely on it on next submissions, thanks !

elieserdejesus:
@fab672000 , I suggested the new lines for brackets, but now after see the result I think is too many new lines for brackets. Honestly I think is better match strictly with Qt code style. With the current format we are exactly in the "wrong" side of Qt code Style 😄 . I vote to move to the "correct" side, and sorry for bother you with these details. I promisse this is the last time!

image

fab672000:
NP at all, we need to be sure before everyone applies this, will change all the braces back to qt style

fab672000:
reverted the braces change can you test and tell me if its enough now or if we should change more?

fab672000:
Thank you! I'll check regurlarly the style on our regularly modified files so that we can progressively get more consistent sources while not generating too many changes just for formatting.

Issue #1036, opened 2018-03-12
[Game theme] Some strings are cropped due to to big font


jonjamcam:
* Looper:

1. "hear all/locked" in polski, deutsch and español:

![image](https://user-images.githubusercontent.com/15310433/37312207-dd32fddc-2628-11e8-98bb-6687455675c0.png)

2. Combobox strings out of range (happens some times only):

![image](https://user-images.githubusercontent.com/15310433/37312376-955dbb36-2629-11e8-9c48-a09a303fa276.png)

*Chat: username when font is at maximum size:

![image](https://user-images.githubusercontent.com/15310433/37312305-43681416-2629-11e8-921e-d62dad43a5f1.png)

jonjamcam:
Btw @elieserdejesus game theme is absolutely amazing. I love it!!!!

elieserdejesus:
> Btw @elieserdejesus game theme is absolutely amazing. I love it!!!!

Thanks @jonjamcam. 

About the font size: 
- I think reduce the font size will make the text unreadable, but I not tried.
- Another solution is search another cool "game" font using less horizontal space.

jonjamcam:
> I think reduce the font size will make the text unreadable, but I not tried.

I just tested that in the case of the looper checkboxes and I think it's not a good solution IMO.

![image](https://user-images.githubusercontent.com/15310433/37434656-0bc92d1e-27bf-11e8-8771-780834f730f3.png)

Before looking for another font I will try resizing the looper window see if it works

elieserdejesus:
@jonjamcam , I'm seeing your image here and the text is huge. I think we need a better translation, not a better font. In general is possible use small words.  

@qulf , can you help suggesting alternative translations using small words?

PR #1156, opened 2018-11-27
adjusting busy dialog size and font to allow long server names

jonjamcam:
Final result:

image

Game theme still crops server name. Added entry to https://github.com/elieserdejesus/JamTaba/issues/1036


elieserdejesus:
thanks @jonjamcam !

PR #1158, opened 2018-11-28
Bulk ip lookup in geolocation

elieserdejesus:
Both geolocation APIs are supporting bulk ip requests. At moment Jamtaba is requesting one IP per HTTP request. This is contributing to "burn" the free API quota.

Using the bulk ip lookup feature we send a set of ips per request and receive a set of geolocations. This can drastically reduce the free APIs usage.


elieserdejesus:
PR canceled to avoid merge conflicts.

PR #1172, opened 2019-01-17
Translations lupdate plus spanish finished

jonjamcam:
@elieserdejesus and @fab672000 while testing this PR I got a strange crash happening in discord server when using the chord creation dialog. Can you test before merge please?

The crash happens as follows:

1.- Launch Jamtaba.
2.- enter discord server
3.- open chord creation dialog and load the default chord.

 after some seconds jamtaba crashes.

If this happens to you guys, then it's probably the es_CL.qm file that's broken. Can you confirm?

EDIT: I confirm It's not this PR's problem, but the crash happens in my system anyway if I compile the master branch. I believe it has something to do with https://github.com/elieserdejesus/JamTaba/commit/00fed62217373f2f930c41695136572bbd247492 because official 2.1.8 is not crashing at all.

EDIT2: The crash I'm pretty sure is related to Discord's Bot and chords. No problems when the bot is offline. Crash occurs when using chords while the bot is on. creating a separate issue https://github.com/elieserdejesus/JamTaba/issues/1173


elieserdejesus:
Thanks for the translation updates @jonjamcam .

I tagged the #1173 for the next release!

PR #1180, opened 2019-02-25
centering country names when hovering map with mouse pointer

jonjamcam:
I think it's cleaner this way if rooms are crowded in small screens or as plugin. Avoiding most overlaps this way.

image

### DIFF
--- a/src/Common/gui/widgets/MapWidget.cpp
+++ b/src/Common/gui/widgets/MapWidget.cpp
@@ -394,7 +394,7 @@ void MapWidget::drawCountriesLegend(QPainter &p)
     auto legendHeight = (flagSize.height() + TEXT_MARGIM) * uniqueCountryMarkers.size() + TEXT_MARGIM;
 
     QSizeF legendSize(legendWidth, legendHeight);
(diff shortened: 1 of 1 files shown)


src/Common/gui/widgets/MapWidget.cpp
```diff
@@ -394,7 +394,7 @@ void MapWidget::drawCountriesLegend(QPainter &p)
     auto legendHeight = (flagSize.height() + TEXT_MARGIM) * uniqueCountryMarkers.size() + TEXT_MARGIM;
 
     QSizeF legendSize(legendWidth, legendHeight);
-    QPointF legendTopLeft(width() - legendSize.width() - TEXT_MARGIM, TEXT_MARGIM);
+    QPointF legendTopLeft((width() - legendSize.width() - TEXT_MARGIM) / 2, TEXT_MARGIM);
```
elieserdejesus: Nice @jonjamcam !

Maybe is better center vertically too. 

Try replace the second parameter (TEXT_MARGIM) by
```
height()/2 - legendSize.height()/2
```

jonjamcam: keeping above syntax too for horizontal parameter:

image

elieserdejesus:
Thanks @jonjamcam !

PR #1195, opened 2020-03-18
Instructions for installing libqt5websockets5

### LINKED ISSUE #1188: libQt5WebSockets.so.5: cannot open shared object file
When trying to launch JamTaba on Linux, I get the following error:

> ./Jamtaba2: error while loading shared libraries: libQt5WebSockets.so.5: cannot open shared object file: No such file or directory

brylie:
Closes #1188


elieserdejesus:
Thanks!

PR #1232, opened 2020-04-03
battery meter (Windows only)

jonjamcam:
Hi @elieserdejesus . I did this little PR to allow battery percentage when portable hardware is used.

 Additionally the memory meter appears only when using more than 60% and battery meter is shown only if a battery is available.

image

The only problem I found was this little empty label when nothing is shown:

image

what do you think?

### DIFF
--- a/src/Common/gui/MainWindow.cpp
+++ b/src/Common/gui/MainWindow.cpp
@@ -2083,17 +2083,25 @@ void MainWindow::timerEvent(QTimerEvent *)
     if (now - lastPerformanceMonitorUpdate >= PERFORMANCE_MONITOR_REFRESH_TIME) {
 
         if (performanceMonitorLabel) {
-            QString string;
 
-                if (performanceMonitor->getMemmoryUsed() > 60) //memory meter only active if memory usage is <60%
-                    string = string + QString("MEM: %1%").arg(performanceMonitor->getMemmoryUsed());
+                   auto memmoryUsed = performanceMonitor->getMemmoryUsed();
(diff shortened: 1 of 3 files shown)


elieserdejesus:
Nice @jonjamcam , working here in win 10.

**About the empty label**: the trick is show the label when one of the 2 conditions is true, or both conditions.

I refatored the code to avoid some duplicated calls in performanceMonitor, but last line is doing the magic in the label visibility (will be visible only when necessary):
```
        if (performanceMonitorLabel) {
      
            auto memmoryUsed = performanceMonitor->getMemmoryUsed();
            auto batteryUsed = performanceMonitor->getBatteryUsed();
            
            bool showMemmory = memmoryUsed > 60; //memory meter only active if memory usage is <60%
            bool showBattery = batteryUsed < 255; //Battery meter active only if battery is available
            
            QString string;
            if (showMemmory) 
                string += QString("MEM: %1%").arg(performanceMonitor->getMemmoryUsed());

            if (showBattery) 
                string += QString(" BAT: %1%").arg(performanceMonitor->getBatteryUsed());

            performanceMonitorLabel->setText(string);
            
            performanceMonitorLabel->setVisible(showMemmory || showBattery);

        }

```

jonjamcam:
>I refatored the code to avoid some duplicated calls in performanceMonitor, but last line is doing the magic in the label visibility (will be visible only when necessary):

cool @elieserdejesus . Just added the refactoring. It's ready for merge. THX!

elieserdejesus:
thanks @jonjamcam , merging!

PR #1243, opened 2020-04-07
Memory is ok for Linux builds (memory thresold "showMemmory" inverted)

arnaudlecam:
Sorry about all those commits,  Git is quite new for me...
I don't know how to include static libs, so I just inclued /sys/sysinfo.h "as is"


src/Common/gui/MainWindow.cpp
```diff
@@ -2087,7 +2087,7 @@ void MainWindow::timerEvent(QTimerEvent *)
                    auto memmoryUsed = performanceMonitor->getMemmoryUsed();
                    auto batteryUsed = performanceMonitor->getBatteryUsed();
 
-                   bool showMemmory = memmoryUsed > 60; //memory meter only active if memory usage is <60%
```
jonjamcam, line 2090: Hi @arnaudlecam I think the comment is wrong here, not the operator. The idea is to show the memory used as a warning when it exceeds 60% . It was fixed before, but when you have lot's of memory available it is useless Sometimes other software or big plugin libraries consume a lot of ram, so this is a quick indicator that something is wrong.

elieserdejesus:
@arnaudlecam , there are many commits, merges, etc in this pull request.

I can`t merge something like that because I need to check everything, it`s a lot of uncessary work.

To create a clean PR your master branch need be in sync with the official master branch (this repository). 

Try
```
git checkout master
git reset --hard origin
git pull
´``
No your master branch is a clone of the remote and you are in sync

After ensure your master branch is in sync Always create a new branch before do your changes
```
git checkout -b cool-branch-name-here
```

PR #1252, opened 2020-04-08
Performance monitoring

arnaudlecam:
Hi @elieserdejesus 
I finally wiped my initial fork, created an new one, and now it works !
- corrects showMemmory's comment in MainWindow.cpp ("memory" replaced by "RAM", showed > 60%) ;
- adds PerformanceMonitor::getMemmoryUsed() for running on Linux (with RAM size) ;
- adds PerformanceMonitor::getBatteryUsed for compiling on Linux (returns 0) ;
- corrects getMemmoryUsed's comment in WindowsPerformanceMonitor.cpp.


elieserdejesus:
Nice @arnaudlecam !

PR #1256, opened 2020-04-11
make voice chat button barely blink when not xmitting

jonjamcam:
- [x] fixing blinking border for Rounded theme
- [x] fixing blinking border for Volcano theme

the idea is to have blinking when actually using the feature and a soft same color blinking when not transmitting. This is just a color match of the blinking color when not transmitting, so it's easy to revert in css


jonjamcam:
@elieserdejesus PR complete!

elieserdejesus:
thanks @jonjamcam !

PR #1261, opened 2020-04-13
MiniUPnP & minimp3 incorporation + dynamic linking for Linux

arnaudlecam:
Compiling for Linux has been a bit tricky (for me...).
In one hand, I propose to integrate MiniUPnP and minimp3, because they are not (or partially) packaged.
In the other hand, I propose to make it possible to dynamically link other dependencies.
This way, perhaps the project would be more "packageable" (it could be nice for Linux users !)...


elieserdejesus:
We have to discuss your proposals. You are including the MiniUpNo and minimp3 directly in the code, but you are doing this not just for Linux (your usage), your changes will affect Windows and Mac.

Are you testing your changes in the 2 plataforms? You must do that before I accept your changes, because if I remember correctly is not too simple compile minimp3 in windows msvc. 

Linux users are just a small part of jamtaba community. Make no sense break windows and Mac versions just to be "linux packageable".  If you discover how be linux packageable without affect Windows and Mac we are ok.

image

arnaudlecam:
Hi @elieserdejesus 
I understand, thank you. It will be closed until I will be able to test it on Windows & Mac.

PR #1265, opened 2020-04-17
fix PortAudioDriver pointer

alessiol78:
i have fixed a null pointer that crash start app with invalid audio config


elieserdejesus:
thanks @alessiol78 !

alessiol78:
I suggest to look this project https://qtractor.sourceforge.io/ because support VST3 plugin and LADSPA also. I build it and I think that it is a good source for new idea. In this days I try to connect your app to QAudioInput class for solve several problems on ALSA, infact your app several times crash for alsa errors and audio config so I try to solve it.

bye

elieserdejesus:
Linux ALSA improvemenets are welcome @alessiol78 .

PR #1334, opened 2020-06-07
fixing crash when user = bot name 'booga'

jonjamcam:
Jamtaba is crashing when user inside room is named as bot 'booga'.

### DIFF
--- a/src/Common/ninjam/client/Service.cpp
+++ b/src/Common/ninjam/client/Service.cpp
@@ -207,7 +207,6 @@ QStringList Service::buildBotNamesList()
     names.append(QStringLiteral("localhost"));
     names.append(QStringLiteral("dojcbot"));
     names.append(QStringLiteral("DOJCbot"));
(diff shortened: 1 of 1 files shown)


elieserdejesus:
@jonjamcam , let me investigate why this name is crashing. I suspect this is a more general problem and other names will crash in future too. Is better solve the problem in the "root". I will check this issue soon.

elieserdejesus:
@jonjamcam , I tested here in local server and can't reproduce the bug. 

I just connected in my local server using 'booga' as username (in ReaNinjam) and connected in the server using JamTAba. No crash!

The problem is only when theres a boot named 'booga' and an user with same name, right?

Any ideas to reproduce the bug?

jonjamcam:
> The problem is only when theres a boot named 'booga' and an user with same name, right?

Yes. It happened to me some days ago. Server was ninjam.live:2049 I think. The user was from US.

I tried to enter several times, the crash happened when the first bpi interval of 120/16 ends

PR #1338, opened 2020-06-10
Adding Saxophone, Flute and Trombone icons

jonjamcam:
- [x] instrument names scrambled

PR finished @elieserdejesus . I think there will be confusion when users update because persistence `"instrument":` is stored as number and the order has changed for some icons with this PR. Some users will find it's local icon changed.


elieserdejesus:
Thank you @jonjamcam !!

> Some users will find it's local icon changed.

Small issue, not a big problem! People will survive an icon change 😄

PR #1357, opened 2020-08-28
match syncCheckbox to Black, Ice and Volcano theme palette

jonjamcam:
image

image

image


elieserdejesus:
thanks @jonjamcam , very nice!!

PR #1363, opened 2020-09-13
Videochannel usage won't xmit audio data on ch2

jonjamcam:
ATM activating camera creates a Videochannel  on channel 2 and simultaneously silent audio data is sent to the server. This means more bandwidth is used. 

Also if you record locally a silent 2nd audio channel is created. This is not optimal.

This PR fixes this issue creating a clean video channel


elieserdejesus:
nice @jonjamcam !

PR #1364, opened 2020-09-14
Allow disable inactivity monitor in json file

jonjamcam:
Allow value = 0 to disable in json file


elieserdejesus:
thanks @jonjamcam !

PR #1370, opened 2020-11-15
bots subtracted from maxusers in jamroomview

jonjamcam:
- [x] change bot name "ninbot" to "ninbot_"

ugly hack but working :/

now all ninbot and ninjamer servers show the correct maxusers value.

TODO: delete obsolete bot detection code.


elieserdejesus:
thanks @jonjamcam

PR #1380, opened 2020-12-29
Fix for Windows crash related to MIDI clock sync

ndonald2:
### Purpose

This is a potential fix for a crash on Windows related to MIDI clock sync output

### Summary

* Wrap message send in a try/catch and disconnect port if an exception is thrown

### Additional Notes

⚠️ **I need help testing this** - I am unable to build Jamtaba in Windows. Although this fix is based on best available information from crash logs, it has not been proven to fix the issue.

**Repro Steps:**

1. Set up MIDI clock sync in the current standalone release version of Jamtaba on Windows, so clock is being sent to a hardware device over USB
2. Join a server or host one locally
3. Disconnect the hardware device that is receiving MIDI clock

This should instantly produce a crash. The fix I've implemented here in theory should prevent the crash. RtMidi does not appear to have any mechanisms for determining if a MIDI output has been disconnected, but if the `sendMessage` fails, it will throw an exception. This will catch the exception and close the port so it will not try to send any more messages in the future. All of the output ports will be reinitialized when preferences are next changed or if Jamtaba is restarted.

### DIFF
--- a/src/Common/midi/RtMidiDriver.cpp
+++ b/src/Common/midi/RtMidiDriver.cpp
@@ -217,8 +217,13 @@ void RtMidiDriver::consumeMessagesFromStream(RtMidiIn *stream, int deviceIndex, 
 
 void RtMidiDriver::sendMessageToOutputs(const std::vector<unsigned char> message) const {
     for (auto stream : midiOutStreams) {
-        if (stream->isPortOpen())
+        if (!stream->isPortOpen()) return;
+        try {
             stream->sendMessage(&message);
+        } catch (RtMidiError &e) {
+            qCritical() << "Error sending midi message to output " << QString::fromStdString(e.getMessage());
+            stream->closePort();
+        }
     }
 }
 


jonjamcam:
hi @ndonald2 . I confirm the crash in 2.1.16. 

 Just tested your branch in windows 7 x64 and the fix is working as expected:

1.- connect keystep and launch JTB standalone.
2.-  connect to server and sync.
3.- disconnect usb cable.

results: No crash.

image

ndonald2:
@jonjamcam wow, amazing! That was so fast! Thanks so much for testing. I've tested on macOS already, it doesn't create any problems, and there was no crash on mac before because of the difference in how CoreMidi handles things under the hood.

ndonald2:
Also as a future heads up: from testing with friends who use Windows, they are seeing clock drift and audio issues occasionally depending on their configuration. It seems Windows MIDI API is synchronous/blocking, so sending the clock messages from inside the audio callback is not ideal especially with multiple clock destinations. It can prevent the buffer callback from finishing in time and clocks/buffers can get dropped. I'm planning on reworking the MIDI clock output strategy at some point but that will be a longer project. For now just saying all this for the sake of transparency.

jonjamcam:
> testing with friends who use Windows, they are seeing clock drift and audio issues occasionally depending on their configuration.

Yes. My experience has been that once in a while the clock sync is lost so the external machine hooked up reverts to internal clock.  I've not use much midi lately, so testing was limited, but if you need help in Windows I'll be glad to help.

elieserdejesus:
thanks @ndonald2  and @jonjamcam !!

PR #1381, opened 2020-12-29
Introduce CI workflow with GitHub actions

Dewb:
Hi! Congrats on the 2.1.16 release, and thanks for providing a way for musicians to collaborate while distancing, it's definitely made 2020 a little easier.

I know that managing build and release pipelines across multiple platforms can be time-consuming, and I've been poking at some possible directions to automate some of the process. This PR is an initial suggestion for some automated build infrastructure. There's definitely more that could be done here, but the current state seems like a good starting point for discussion.

What merging this PR in its current state would accomplish:
- Future PRs would get "build checks" to confirm each new commit builds on all three platforms (Windows, macOS, Linux)
   image
- The artifacts from these builds will be downloadable from the Actions tab, so people could test PRs on different platforms before merging the PR
   image
- Changes to the master branch will also produce builds for all three platforms in the Actions tab

What this does not yet do, but could with a little work/discussion:
- Build the Windows installer (Need to capture and automate which tools/steps are required)
- Compile the VST project on Windows (Linux works, but there are compilation errors in the Windows build for some reason)
- Compile the AU project on macOS (compilation errors)
- Post prerelease-tagged installers for all platforms on the Releases page when new PRs are merged

What this could be further extended to do in the future:
- Build the binary dependencies too (portaudio, libogg, etc.), to make it easier to adopt new toolchains/compilers (e.g. a newer version of Visual Studio)
- Test against newer versions of Qt
- Make it easier to add support for additional platforms/architectures (M1 Macs?)

### DIFF
--- a/installers/linux/build_installer.sh
+++ b/installers/linux/build_installer.sh
@@ -81,7 +81,7 @@ cp $qtLibDir/libQt5XcbQpa.so.* packageFiles/
 cp $qtLibDir/libicui18n.so.* packageFiles/
 cp $qtLibDir/libicuuc.so.* packageFiles/
 cp $qtLibDir/libicudata.so.* packageFiles/
-cp $qtLibDir/qt5/plugins/platforms/libqxcb.so packageFiles/platforms
+cp $qtDir/plugins/platforms/libqxcb.so packageFiles/platforms
 
 
 chmod +x packageFiles/installer_script.sh 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,191 @@
+name: CI Build
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build-mac:
+    runs-on: macos-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Qt
+        uses: jurplel/install-qt-action@v2
+        with:
+          version: '5.9'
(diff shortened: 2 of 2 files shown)


installers/linux/build_installer.sh
```diff
@@ -81,7 +81,7 @@ cp $qtLibDir/libQt5XcbQpa.so.* packageFiles/
 cp $qtLibDir/libicui18n.so.* packageFiles/
 cp $qtLibDir/libicuuc.so.* packageFiles/
 cp $qtLibDir/libicudata.so.* packageFiles/
-cp $qtLibDir/qt5/plugins/platforms/libqxcb.so packageFiles/platforms
+cp $qtDir/plugins/platforms/libqxcb.so packageFiles/platforms
```
Dewb, line 84: This path may be dependent on the Qt version, I had to change it to build with 5.9.

elieserdejesus:
Hi, the big problem to compile JamTaba using CI is build Qt statically. We need Qt static builds for VST and AU plugins. 

The Qt static builds are not officially tested, so is always complicated find the Qt static compilable version. I tested a lot of versions and have 99% of compilation errors. It`s a very time consuming task find a static compilable version.

Sometimes a Qt version compile statically in Windows but fails in Mac 😢 

Another approach is try to find a pre compiled Qt (static and working in Windows & Mac) and download this Qt version using the CI scripts. But again it`s a very time consuming task.

PR #1382, opened 2020-12-29
updating maxusers hack

jonjamcam:
this PR fixes maxusers for some servers that use hidden bots such as discordonlinejammingcentral and mutant.


elieserdejesus:
thanks @jonjamcam !

PR #1409, opened 2021-10-30
update public view hack for correct room capacity

jonjamcam:


ifxsoftware:
I just spent some time fixing the live serverlist....
user_max  -  now correct  
user_count - added and is from the live servers so includes bots

zenjam:
I just spent some time fixing the live serverlist....
user_max - now correct
user_count - added and is from the live servers so includes bots

zenjam:
Okay correcting the user_max caused jamtaba to have the wrong number
So ive added user_limit to show the correct number for max users
and added one for user_max jamataba 

Summary please feel free to user "user_limit" for future versions and
user_max will be left for older jamtaba versions.

-Tom

jonjamcam:
Hi Tom, Elieser.  I find the updated user_max value is showing +1 value
compared to cockos list. But it's foing to get obsolete as soon as the new
version is released.

Just tested the new 'user_limit' entry. This value is the correct one for
most servers, except for ninbot and ninjamer (considering yesterday's
commit using the updated hack).

For this reason if Tom is using existing official releases to address this
issue, the value will always be +1 for ninbot and ninjamer servers.

Happy to work on this if needed.


cheers

zenjam:
I think the confusion is that the bot names are removed from the users list
so they wouldnt show up in serverlist

We need to sort this out maybe
user_count includes the bot
user_limit is how many can join including the bot
user_max is user_limit - 1  (to correct for current jamtaba)

jonjamcam:
I think that's the issue too

